PR #1, opened 2017-05-21
Code reorganization

slide:
There are some test failures that don't occur in "main" and not all of the same tests are being run yet.


slozier:
Looks like a typo in assert_util.py which is causing most of the failures? I didn't have time to test with the typo fixed.

I wonder if we should wait until after the next release to see how much of an outcry there is before we start ripping out all the silverlight and net35 stuff. Although I guess it's fine to remove it from testing since we're not running those tests anyway...

slozier:
Maybe move the test stuff into a separate PR?

slide:
Ok, I can do that, my reason for putting it here was because I didn't want to fix the tests in the current state, I wanted to move them to the new format while fixing them.

slozier:
Only reason I'm suggesting using a separate PR is to avoid having the changes getting lost in the noise (we'll need to port them over to ipy3). Once the testing changes are split off we can probably merge in the reorg as is.

slide:
Good point

PR #2, opened 2017-05-25
[WIP] Unit test updates

slide:
Updates to IronPython specific unit tests to use the unittest


Tests/test_builtinfunc.py
```diff
@@ -21,718 +24,716 @@
...
-def cp946():
-    builtins = __builtins__
-    if type(builtins) is type(sys):
-        builtins = builtins.__dict__
-    
-    if "hasattr" not in builtins:
-        raise "hasattr should be in __builtins__"
-    if "HasAttr" in builtins:
-        raise "HasAttr should not be in __builtins__"
+class BuiltinsTest1(unittest.TestCase):
```
slozier: According to the comment these should be run before importing iptest.assert_util. Do we have any idea what the issue was and if using unittest prevents these from working properly?

Tests/test_builtinfunc.py
```diff
@@ -21,718 +24,716 @@
...
+        
+        self.assertIn('hasattr', builtins, 'hasattr should be in __builtins__')
+        self.assertNotIn('HasAttr', builtins, 'HasAttr should not be in __builtins__')
 
-cp946()
-if sys.platform=="cli":
-    import clr
-cp946()
+    def test_no_clr_import(self):
+        self.cp946()
```
slozier: Does unittest guarantee that this one will run before the next one?

Tests/test_execfile.py
```diff
@@ -13,32 +13,31 @@
...
-
-def test_scope():
-    root = testpath.public_testdir
-    z = 10
-    execfile(path_combine(root, "Inc", "execfile_scope.py"))
+import unittest
+
+from iptest import IronPythonTestCase, is_cli
+
+@unittest.skipUnless(is_cli, 'IronPython specific test case')
```
slozier: The skip was only on test_sanity. Although unless there's something cli specific in toexec.py it might not be necessary.

Tests/test_excinfo.py
```diff
@@ -1,8 +1,8 @@
-#####################################################################################
+##########################################################################
 #
 #  Copyright (c) Microsoft Corporation. All rights reserved.
 #
-# This source code is subject to terms and conditions of the Apache License, Version 2.0. A
+# This source code is subject to terms and conditions of the Apache License, Version 2.0. self.A
```
slozier: Added a typo?

Tests/test_exceptionconverter.py
```diff
@@ -1,142 +0,0 @@
-#####################################################################################
```
slozier: Where did this file one go?

Tests/test_attribute_customize.py
```diff
@@ -28,245 +29,249 @@ def return_100(self, name):        return 100
...
-    class D(C1): pass
-    d = D()
-    d.x = 20
-    AreEqual([d.x, d.y], [30, 100])
-    
-    C1.__getattribute__ = lambda *args: 200  # __getattribute__ not honored
-    class D(C1): pass
-    d = D()
-    AreEqual([d.x, d.y], [100, 100])
+    def test_delete_getattribute(self):
```
slozier: This one was disabled. Is it fixed now?

Src/IronPython/Lib/iptest/assert_util.py
```diff
@@ -72,68 +72,52 @@ def remove_ironpython_dlls(targetdir):
...
-        ipython_executable  = path_combine(sys.prefix, 'ipy.exe')
-        cpython_executable  = sys.executable
-    
-    #team_dir            = path_combine(ip_root, r'Team')
-    #team_profile        = path_combine(team_dir, r'settings.py')
-    #
-    #my_name             = os.environ.get(r'USERNAME', None)
-    #my_dir              = my_name and path_combine(team_dir, my_name) or None
-    #my_profile          = my_dir and path_combine(my_dir, r'settings.py') or None
+#     #team_dir            = path_combine(ip_root, r'Team')
```
slide: No, it's part of what makes this WIP still...I need it to refer back to as I work through test. It will be removed in the final commit.

Tests/Inc/execfile_scope.py
```diff
@@ -13,4 +13,4 @@
 #
 #####################################################################################
 
-AreEqual(z, 10)
+self.assertEqual(z, 10)
```
slide: No, this is imported in test_execfile.py, it's just used for testing.

Tests/Inc/toexec.py
```diff
@@ -13,8 +13,6 @@
 #
 #####################################################################################
 
-from iptest.assert_util import *
-
-Assert(__name__ in  ["__main__", "test_execfile"], __name__)
+self.assertTrue(__name__ in  ["__main__", "test_execfile"], __name__)
```
slide: No, this is imported in test_execfile.py

Tests/test_attr.py
```diff
@@ -13,283 +13,283 @@
...
+        # hasattr shouldn't swallow SystemExit exceptions.
+        class x(object):
+            def __getattr__(self, name):
+                import sys
+                sys.exit(1)
+        
+        a = x()
+        try:
+            hasattr(a, 'abc')
+            AssertUnreachable()
```
slide: Yes, when it works I don't see the errors for wrong assertion methods :-)

Tests/test_clrload2.py
```diff
@@ -14,46 +14,51 @@
...
-        obj, name = x[0], x[1]
-        
-        AreEqual(getattr(obj, 'CallMe' + name)(), name + ' Hello World')
-        
-        AreEqual(getattr(obj, name+'Field'), None)
+import unittest
+
+from iptest import IronPythonTestCase, is_netstandard, is_mono
+
+class ClrLoad2Test(IronPythonTestCase):
```
slide: Yes, good catch

Tests/test_builtinfunc.py
slide: No, I don't know what the issue was. Maybe need to revisit it.

Tests/test_builtinfunc.py
slide: No, but isn't the import local to the method?

Tests/test_execfile.py
slide: You are right, I don't think it's needed at all.

Tests/test_builtinfunc.py
slozier: Yes the clr variable is local to the function, but just fact that we're importing clr causes some side effects (such as the C# methods becoming visible). For example:
```Python
def test():
    import clr
assert "ToString" not in dir("a")
test()
assert "ToString" in dir("a")
```

slide: I'll rename the methods, there is this note in the unittest docs:

"Note that the order in which the various test cases will be run is determined by sorting the test function names with respect to the built-in ordering for strings."

Tests/test_attribute_customize.py
slide: Yes, it runs correctly. I generally have been enabling all the tests and then looking through to see if there really is something that should be disabled. I am also trying to use expectedFailure instead of disabling if there is an issue, that way it is noted in the logs.

Tests/test_is.py
```diff
@@ -13,36 +13,41 @@
...
-from iptest.assert_util import *
+import unittest
 
-from System import Nullable
+from iptest import is_cli
 
-def test_object():
-    a = object()
-    b = object()
+@unittest.skipUnless(is_cli, 'IronPython specific test case')
```
slozier: Move the skipUnless over to test_bool_nullablebool? test_object should work with CPython.

Tests/test_isinstance.py
```diff
@@ -13,1072 +13,1013 @@
...
-del pow
-AreEqual(operator.isCallable(pow), True)
-
-try:
-    del pow
-    AreEqual(True,False)
-except NameError:
-    pass
-
-def test_file():
```
slozier: Dropped this test?

Tests/test_listcomp.py
```diff
@@ -13,65 +13,50 @@
...
+
+        self.assertRaises(NameError, lambda: [z for x in "python"])
+        self.assertRaises(NameError, lambda: [x for x in "python" if z > 5])
+        self.assertRaises(NameError, lambda: [x for x in "iron" if z > x for z in "python" ])
+        self.assertRaises(NameError, lambda: [x for x in "iron" if never_shown_before > x ])
+        self.assertRaises(NameError, lambda: [(x, z) for x in "iron" if z > x for z in "python" ])
+        self.assertRaises(NameError, lambda: [(i, j) for i in range(10) if j < 'c' for j in ['a', 'b', 'c'] if i % 3 == 0])
+
+    def test_negative_flow_checker(self):
+        self.assertRaises(NameError, lambda: [y for x in "python"])
```
slozier: This seems like exactly the same as the test above?

Tests/test_property.py
```diff
@@ -13,233 +13,242 @@
...
+        if is_cli:
+            self.array_list_options.append(System.Collections.Generic.List[int])
+            if is_netstandard:
+                import clr
+                clr.AddReference("System.Collections.NonGeneric")
+            self.array_list_options.append(System.Collections.ArrayList)
+
+    def test_stdin(self):
+        self.assertTrue('<stdin>' in str(sys.stdin))
+        #self.assertTrue('<stdout>' in str(sys.stdout))
```
slozier: Can we uncomment these?

Tests/test_str.py
```diff
@@ -14,439 +14,434 @@
...
-    AreEqual(id(x), id(y))
-    AreEqual(id(x), id(False))
-    
-    x = 'abcx' != 'abc'
-    y = 'defx' != 'def'
-    AreEqual(id(x), id(y))
-    AreEqual(id(x), id(True))
-
-def test_str_dict():
-    print "CodePlex Work Item 13115"
```
slozier: Maybe preverse the codeplex # as a comment?

Tests/interop/net/field/test_field_misc.py
```diff
@@ -16,28 +16,31 @@
...
 
+import unittest
 
-add_clr_assemblies("fieldtests", "typesamples")
+from iptest import IronPythonTestCase
 
-if options.RUN_TESTS: #TODO - bug when generating Pydoc
-    from Merlin.Testing.FieldTest import *
-    from Merlin.Testing.TypeSample import *
+class FieldMiscTest(IronPythonTestCase):
```
slozier: We probably want skipunless is_cli on all of these.

Tests/test_assert.py
```diff
@@ -16,68 +16,75 @@
...
+        self.assertTrue(ok)
   
   
 #--Main------------------------------------------------------------------------
-if is_cli and '-O' in System.Environment.GetCommandLineArgs():
-    from iptest.process_util import *
-    AreEqual(0, launch_ironpython_changing_extensions(__file__, remove=["-O"]))
-else:
-    run_test(__name__)
+# if is_cli and '-O' in System.Environment.GetCommandLineArgs():
```
slide: It doesn't quite do what was originally there, but we could add the check for optimized.

Tests/test_exceptionconverter.py
slide: No tests were being run from this file and the framework for running the test no longer exists in the testing libraries.

Tests/test_file.py
```diff
@@ -13,827 +13,831 @@
...
+
+        # Actually run the pattern mode tests.
+        test_patterns()
+
+    # Now some tests of read(size).
+    # Test data is in the following format: ("raw data", read_size, (binary mode result strings) (binary mode result tell() result)
+    #                                                               (text mode result strings) (text mode result tell() result)
+    #                                                               (universal mode result strings) (universal mode result tell() results)
+
+    @unittest.skipUnless(is_cli, 'IronPython specific test')
```
slide: No, I just maintained what is there on this one. I don't see us running most of these tests under CPython, but I guess its possible.

Tests/test_imp.py
```diff
@@ -13,1344 +13,1318 @@
...
-    imp.init_builtin("exceptions")
-
-@skip("win32", "multiple_execute", "stdlib")
-def test_sys_path_none_builtins():
-    prevPath = sys.path
-
-    #import some builtin modules not previously imported
-    try:
-        sys.path = [None] + prevPath
-        Assert('datetime' not in sys.modules.keys())
```
slide: Yes, unittest will import datetime (at some level), so datetime is in sys.modules since we are now using unittest.

Tests/test_isinstance.py
slide: Yes, it didn't seem to fit with the test of the tests.

Tests/test_listcomp.py
slide: Yeah, I'll remove the second one.

Tests/interop/net/field/test_field_misc.py
slide: I thought of that too, but wasn't sure we would ever try running these under CPython.

slozier: I'm actually thinking about running the tests under CPython. Maybe not under CI, but it would be useful to highlight compatibility issues.

Tests/test_isinstance.py
```diff
@@ -234,6 +233,7 @@ def __getitem__(self, index):
 
     def test_eval_dicts(self):
         # eval referencing locals / globals
+        import System
```
slozier: Maybe move the import to the if is_cli block?

Tests/test_isinstance.py
```diff
@@ -1006,20 +1006,6 @@ def test_tuple_new(self):
         self.assertRaises(TypeError, tuple.__new__, str)
         self.assertRaises(TypeError, tuple.__new__, str, 'abc')
 
-    def test_override_pow(self):
```
slozier: Something wrong with this test?

Tests/interop/net/test_loadorder.py
```diff
@@ -1,125 +0,0 @@
-#####################################################################################
```
slozier: Why remove this test?

Tests/interop/net/type/test_reachtype.py
```diff
@@ -189,65 +191,67 @@ def test_type_from_reflection_emit():
...
-    add_clr_assemblies("typeforwarder1")
-    from NSwForwardee1 import Foo, Bar        #!!!
-    AreEqual(Foo.A, 120)
-    AreEqual(Bar.A, -120)
-    
-    import NSwForwardee1
-    AreEqual(NSwForwardee1.Foo.A, 120)
-    AreEqual(NSwForwardee1.Bar.A, -120)
-
-@skip("multiple_execute", "posix")
```
slozier: In other places you kept the multiple_execute as a TODO however it seems you zapped it here.

Tests/modules/io_related/test_binascii.py
```diff
@@ -0,0 +1,60 @@
...
+
+import binascii
+import unittest
+
+from iptest import skipUnlessIronPython
+
+
+class BinasciiTest(unittest.TestCase):
+
+    @skipUnlessIronPython
```
slozier: I think we can remove the skipUnless, seems to work with CPython.

Tests/modules/io_related/cPickle_test.py
```diff
@@ -1,944 +0,0 @@
-#####################################################################################
```
slozier: Why remove this one?

Tests/modules/io_related/test_codecs.py
```diff
@@ -0,0 +1,551 @@
...
+        #--Test that bogus encodings fail properly
+        t_in, t_out, t_err = os.popen3(sys.executable + " " + os.path.join(self.test_dir, "encoded_files", "cp1019.py"))
+        t_err_lines = t_err.readlines()
+        t_out_lines = t_out.readlines()
+        t_err.close()
+        t_out.close()
+        t_in.close()
+        
+        self.assertEqual(len(t_out_lines), 0)
+        print t_err_lines[0]
```
slozier: Are these prints for testing?

Tests/interop/net/test_loadorder.py
slide: It was just calling the other loadorder tests, which will be run automatically now.

Tests/modules/io_related/cPickle_test.py
slide: It's renamed in a coming commit.

Tests/modules/io_related/test_codecs.py
slide: Yes, will remove

Tests/interop/net/type/test_reachtype.py
slide: Should be fixed now

Src/IronPython.Modules/socket.cs
```diff
@@ -2862,23 +2862,23 @@ static ssl()
...
+            [Documentation(@"read([len]) -> string
+
+Read up to len bytes from the SSL socket.")]
+            public string read(CodeContext/*!*/ context, [DefaultParameterValue(Int32.MaxValue)] int len) {
                 EnsureSslStream(true);
 
                 try {
                     byte[] buffer = new byte[2048];
-                    MemoryStream result = new MemoryStream(n);
+                    MemoryStream result = new MemoryStream(len);
```
slozier: Unrelated to the change, but wow, we're really allocating a 2gb array if len is not specified?

Tests/modules/network_related/_ssl_test.py
```diff
@@ -1,385 +0,0 @@
...
-    if is_cpython:
-        AreEqual(str(real_ssl.SSLType),
-                 "<type 'ssl.SSLContext'>")
-    else:
-        #http://ironpython.codeplex.com/WorkItem/View.aspx?WorkItemId=24266
-        AreEqual(str(real_ssl.SSLType),
-                 "<type 'socket.ssl'>")
-                 
-    #--Negative
-    if is_cpython:
```
slozier: Dropped the negative case?

Tests/modules/network_related/test__ssl.py
```diff
@@ -0,0 +1,275 @@
...
+            [DefaultParameterValue(null)] string certfile)
+        '''
+        #--Positive
+        
+        #sock
+        s = socket.socket(socket.AF_INET)
+        s.connect((SSL_URL, SSL_PORT))
+        ssl_s = real_ssl.sslwrap(s._sock, False)
+
+        if not is_cli:
```
slozier: Any reason why the shutdown is excluded here but not in other tests?

Tests/modules/network_related/test__socket.py
```diff
@@ -0,0 +1,606 @@
...
+        missing = [ x for x in expected_dir if x not in fileobject_dir ]
+        self.assertEqual([], missing)
+
+    @retryOnFailure
+    def test_misc(self):
+        f = _socket.socket().makefile()
+        f.bufsize = 4096
+        self.assertEqual(4096, f.bufsize)
+
+    @unittest.skipIf(is_posix, 'some issue on mono, need to debug further')
```
slozier: is_mono?

Tests/modules/system_related/test_nt.py
```diff
@@ -0,0 +1,1174 @@
...
+        self.assertEqual(nt.listdir(nt.getcwd()).count('dir_create_test'), 1)
+        
+        nt.rmdir('dir_create_test')
+        self.assertEqual(nt.listdir(nt.getcwd()).count('dir_create_test'), 0)
+
+    def test_mkdir_negative(self):
+        nt.mkdir("dir_create_test")
+        try:
+            nt.mkdir("dir_create_test")
+            self.assertTrueUnreachabale("Cannot create the same directory twice")
```
slozier: assertTrueUnreachable?

Tests/modules/network_related/_ssl_test.py
slide: No, the file was renamed to match the test_*.py naming convention

Tests/modules/network_related/test__ssl.py
slide: No, I'll fix it

Tests/modules/network_related/test__socket.py
slide: Right, good catch.

Tests/modules/system_related/test_nt.py
slide: replace all issue, will fix

Src/IronPython.Modules/socket.cs
slide: Yes, that is indeed bad...

Tests/modules/network_related/_ssl_test.py
slozier: I meant the positive case of test_SSLType appears in test__ssl.py but I don't see the negative case...

slide: Ah, that's because there is no SSLType for the _ssl module anymore. Both IronPython and CPython get a message saying the module object has no attribute SSLType, so the test was bogus.

Tests/test_kwarg.py
```diff
@@ -152,614 +136,562 @@ def __new__(cls, a, *param, **kw):
...
+
+    def test_negTestFunc_testFunc_pw_kw_dupArg(self):
+        self.assertRaises(TypeError, testFunc_pw_kw, '234', a='234')
+
+    def test_negTestFunc_testFunc_kw_dupArg(self):
+        self.assertRaises(TypeError, testFunc_kw, '234', a='234')
+
+    def test_negTestFunc_ObjectSubClass_testFunc_pw_kw_dupArg(self):
+        o = ObjectSubClass()
+        try: o.testFunc_pw_kw(a='abc')
```
slozier: Any idea why assertRaises doesn't work for these? Maybe add a comment that assertRaises should not be used here?

slide: I think it may have something to do with how the parameter is passed around (in the case of NOT using assertRaises, it is setting the value of the 'a' parameter, but when passed into assertRaises it is passed as **kwArgs. A comment is a good idea

slide: C Python doesn't seem to have any issues with it though, so I guess we have a bug.

PR #9, opened 2017-06-04
More test fixes

slide:


Src/IronPython/Hosting/PythonCommandLine.cs
```diff
@@ -291,12 +291,10 @@ public sealed class PythonCommandLine : CommandLine {
         private void InitializeExtensionDLLs() {
             string dir = Path.Combine(PythonContext.InitialPrefix, "DLLs");
             if (Directory.Exists(dir)) {
-                foreach (string file in Directory.EnumerateFiles(dir)) {
-                    if (file.ToLower().EndsWith(".dll")) {
-                        try {
-                            ClrModule.AddReference(PythonContext.SharedContext, new FileInfo(file).Name);
-                        } catch {
-                        }
+                foreach (string file in Directory.EnumerateFiles(dir, "*.dll")) {
```
slozier: Note that with this change we might also match non-DLL files. The wildcard is actually equivalent to \*.dll\* so you could end up matching other stuff (e.g. a.dllb).

slide: I hadn't meant to push this change. I'll revert

Tests/test_generator_throw.py
```diff
@@ -11,508 +11,520 @@
...
+    #not sure why this is failing when it's no in the global scope
+    @unittest.expectedFailure
+    def test_yield_lambda(self):
+        """
+        Yield can appear in lambda expressions (or any function body).
+        A generator lambda expression yields its final result, instead of just
+        returning it.
+        """
+
+        def test(arg):
```
slozier: If the goal here is to test yield in a lambda expression, why rewrite it as a function?

slide:
I didn't, the test function was already there as a different name.

Tests/test_namebinding.py
```diff
@@ -13,1129 +13,1130 @@
...
+    def test_runtime_name_lookup_class_scopes(self):
+        # Test that run time name lookup skips over class scopes
+        # (because class variables aren't implicitly accessible inside member functions)
+        a = 123
+        class C(object):
+            a = 456
+            def foo(self):
+                # this exec statement is here to cause the "a" in the next statement
+                # to have its name looked up at run time.
+                exec "dummy = 10" in {}, locals()
```
slozier: Does adding ` in {}, locals()` change the behavior of the test?

Tests/test_generator_throw.py
slide: Oh, it looks like it did get changed, I wonder if autopep8 did that...I don't remember changing it, but its all a blur of python code at this point :-)

Tests/test_namebinding.py
slide: I had a concern about that too. I don't know the answer right now.

Tests/test_generator_throw.py
slide: Ok, it looks like autopep8 DOES make this change: https://www.python.org/dev/peps/pep-0008/#programming-recommendations 

I'm not completely crazy after all.

Tests/test_delegate2.py
```diff
@@ -0,0 +1,887 @@
+#####################################################################################
```
slozier: Any reason for the file rename?

slide: There is another test_delegate.py in another directory.

slozier: We should probably change the test runner to handle duplicate name (maybe append the dir structure to the name). I'm pretty sure this isn't the only place this occurs.

Tests/test_generator_throw.py
```diff
@@ -102,32 +102,29 @@ def ff3(l):
         nested()
         gc.collect()
         # in controlled environment like this, this is ok to expect finalizer to run
-        # however, when gc happens at random, and finalizer tries to continue execution
+        # however, when gc happenus at random, and finalizer tries to continue execution
```
slozier: Added a typo here?

slide: Thanks, will fix

Tests/test_stdconsole.py
```diff
@@ -27,6 +27,28 @@
...
+    def __enter__(self):
+        from System import Environment
+        if self._prepend:
+            Environment.SetEnvironmentVariable(self._variable, "%s%s%s" % (self._value, self._sep, self._oldval))
+        else:
+            Environment.SetEnvironmentVariable(self._variable, self._value)
+
+    def __exit__(self, *args):
+        from System import Environment
+        if self._oldval:
```
slozier: Why the check? Wouldn't you want to remove the environment variable if it was not defined before?

PR #31, opened 2017-06-07
Fixes for mono

slide:


Build/net45.props
```diff
@@ -72,7 +72,7 @@
     <Features>$(Features);FEATURE_ASYNC</Features>
     <Features>$(Features);FEATURE_GET_TYPE_INFO</Features>
     <Features>$(Features);FEATURE_DYNAMIC_EXPRESSION_VISITOR</Features>
-    <Features>$(Features);FEATURE_ASSEMBLYBUILDER_DEFINEDYNAMICASSEMBLY</Features>
+    <Features Condition=" '$(Mono)' != 'True' ">$(Features);FEATURE_ASSEMBLYBUILDER_DEFINEDYNAMICASSEMBLY</Features>
```
slozier: Why exclude this? It should be available in Mono.

slide: It was not building for me under 4.8

slozier: Odd. Ah well, it's just a switch between two APIs that are equivalent anyway, we could probably completely remove it from net45.props.

slide:
I'll double check to make sure my codebase wasn't messed up somehow

Build/net45.props
slide: I'll double check to make sure my codebase wasn't messed up somehow

PR #149, opened 2017-06-10
Update to nunit 3.7.1

slozier:


slozier:
@slide So it looks like in the newer NUnit using the static Assert.* methods outside of NUnit tests no longer works as an assert check. Any ideas? Just revert the asserts to whatever we had before?

slide:
Can we turn them into nunit tests?

slozier:
Probably, as far as I can tell test_engine (which is the failing test) is just a launcher to run the C# method. There's also part of test_ipye which does something similar but it looks like it hasn't been updated in your test cleanup.

slide:
If we can make them nunit tests, it would be better

slozier:
Well that test conversion was pleasantly simple to do. Not sure why Travis is failing. The log says everything passed...

slozier:
Still a few issues with this PR:
- the test names of the IronPython category aren't being output
- Travis tests aren't working properly (Linux fails even though there's no failed test, Mac passes even with a failed test)

slozier:
Assuming this passes it should be good to go.

Prep work was done in: IronLanguages/ironpython2#179 and IronLanguages/ironpython2#180

PR #157, opened 2017-06-18
Fix #154

### LINKED ISSUE #154: Can create sets with duplicate elements using pop
From #152. Here's a simplified repro:
```Python
s = set([1,17])
assert s.pop() == 1
s.add(17)
assert tuple(s) == (17,)
```

slozier:
I think this should do the trick though I'm not an expert on how to properly implement hash sets. I'm wondering if we could just switch to using HashSet from the BCL...


slozier:
Closing & reopening to test.

PR #158, opened 2017-06-18
Fix test creating empty folders

slozier:
Changes basename to dirname (the original file had Path.GetDirectoryName).


slide:
Good catch!

PR #159, opened 2017-06-18
Fix #129

### LINKED ISSUE #129: Import modules from a zip archive containing non-ASCII characters does not work
_From @ironpythonbot on December 9, 2014 18:10_

Example module(encoding windows-1251):  

``` python
# coding=windows-1251  
print "Привет мир!"  
```

If import the module from the zip file does not print "Привет мир!"  
## Work Item Details

**Original CodePlex Issue:**    [Issue 34571](http://ironpython.codeplex.com/workitem/34571)
**Status:** Proposed
**Reason Closed:**  Unassigned
**Assigned to:**    Unassigned
**Reported on:**    Oct 17, 2013 at 5:22 PM
**Reported by:**    vabe
**Updated on:** Nov 8, 2013 at 8:49 PM
**Updated by:** vabe
#### Plaintext Attachments

[CodePlex Issue #34571 Plain Text Attachments](https://gist.github.com/c480888d6789270c8801)

_Copied from original issue: IronLanguages/main#1084_

slide:


Src/IronPython.Modules/zipimport.cs
```diff
@@ -212,13 +212,13 @@ private enum ModuleStatus {
...
                 ScriptCode script = null;
-                string code = GetModuleCode(context, fullname, out ispackage, out modpath);
+                byte[] code = GetModuleCode(context, fullname, out ispackage, out modpath);
                 if (code == null) {
                     return null;
                 }
 
                 mod = pythonContext.CompileModule(modpath, fullname,
-                    new SourceUnit(pythonContext, new SourceStringContentProvider(code), modpath, SourceCodeKind.File),
+                    new SourceUnit(pythonContext, new MemoryStreamContentProvider(context.GetPythonContext(), code, modpath), modpath, SourceCodeKind.File),
```
slozier: Hmm, another way to get the language context? Didn't know about this one. Maybe use `pythonContext` instead since we already have it.

slide: Good call, will update

PR #163, opened 2017-06-19
Fix #109

### LINKED ISSUE #109: Expected int got slice, when using start index in for-loop
_From @simplicbe on December 2, 2015 9:23_

When i execute the current script:

```
from System.Collections.Generic import List
lst = List[str]()
lst.Add("Hello")
lst.Add("World")

for i in lst[1:]:
    print i
```

the error message `TypeError: expected int, got slice` is thrown.

If i make a python `list` out of the `lst`, every thing works:

```
for i in list(lst)[1:]:
    print i
```

I try to find the issue, but may be some one has a good idea why this behaviour is as it is.

_Copied from original issue: IronLanguages/main#1251_

slide:
Add methods to make List<T> more pythonic

### DIFF
--- a/Src/IronPython/Runtime/Operations/ListOfTOps.cs
+++ b/Src/IronPython/Runtime/Operations/ListOfTOps.cs
@@ -13,6 +13,8 @@
  *
  * ***************************************************************************/
 
+using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.CompilerServices;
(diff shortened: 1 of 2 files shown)


Src/IronPython/Runtime/Operations/ListOfTOps.cs
```diff
@@ -53,5 +55,70 @@ public static class ListOfTOps<T> {
                 infinite.RemoveAt(index);
             }
         }
+
+        #region Python __ methods
+
+        public static bool __contains__(List<T> l, T item) {
```
slozier: Aren't these three methods already generated automatically?

slide: Not sure, I'll have to check

slozier: I think they're generated by some of the `MemberResolver`s defined in `PythonTypeInfo`

Src/IronPython/Runtime/Operations/ListOfTOps.cs
```diff
@@ -53,5 +57,120 @@ public static class ListOfTOps<T> {
...
+
+        [SpecialName]
+        public static void DeleteItem(List<T> l, object index) {
+            DeleteItem(l, Converter.ConvertToIndex(index));
+        }
+
+        [SpecialName]
+        public static void DeleteItem(List<T> l, Slice slice) {
+            if (slice == null) {
+                throw PythonOps.TypeError("list indices must be integers or slices");
```
slozier: List\<T> to be consistent.

Src/IronPython/Runtime/Operations/ListOfTOps.cs
```diff
@@ -53,5 +57,120 @@ public static class ListOfTOps<T> {
...
+            }
+
+            lock (l) {
```
slozier: Probably don't need the lock here.

slozier:
Close and reopen to get a pass (test_urllib failed).

PR #166, opened 2017-06-19
Adding test for #129

slide:


Tests/test_zipimport.py
```diff
@@ -0,0 +1,29 @@
+#####################################################################################
+#
+#  Copyright (c) Microsoft Corporation. All rights reserved.
```
slozier: We should probably come up with a non-Microsoft copyright notice for new stuff...

slide: We'll get one when we join the .NET Foundation.

PR #173, opened 2017-06-21
Fix #10

### LINKED ISSUE #10: Message for trying to raise 'str' inconsistent with CPython
CPython
```
>>> try:
...     raise 'foo'
... except TypeError, e:
...     print e.message
...
exceptions must be old-style classes or derived from BaseException, not str
```

IronPython
```
>>> try:
...     raise 'foo'
... except TypeError, e:
...     print e.message
...
exceptions must be classes, or instances, not str
```

slide:
Update exception message

### DIFF
--- a/Src/IronPython/Runtime/Generator.cs
+++ b/Src/IronPython/Runtime/Generator.cs
@@ -527,7 +527,7 @@ namespace IronPython.Runtime {
 
             // This may invoke user code such as __init__, thus MakeException may throw. 
             // Since this is invoked from the generator's body, the generator can catch this exception. 
-            return LightExceptions.Throw(PythonOps.MakeExceptionWorker(Context, throwableBackup[0], throwableBackup[1], throwableBackup[2], false, true));
+            return LightExceptions.Throw(PythonOps.MakeExceptionForGenerator(Context, throwableBackup[0], throwableBackup[1], throwableBackup[2]));
         }
 
         private void Close() {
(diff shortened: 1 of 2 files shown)


slide:
Looks like the generator throw needs an update, I'll add it to this PR

Src/IronPython/Runtime/Generator.cs
```diff
@@ -527,7 +527,7 @@ public sealed class PythonGenerator : IEnumerator, IEnumerator<object>, ICodeFor
 
             // This may invoke user code such as __init__, thus MakeException may throw. 
             // Since this is invoked from the generator's body, the generator can catch this exception. 
-            return LightExceptions.Throw(PythonOps.MakeException(Context, throwableBackup[0], throwableBackup[1], throwableBackup[2]));
+            return LightExceptions.Throw(PythonOps.MakeExceptionWorker(Context, throwableBackup[0], throwableBackup[1], throwableBackup[2], false, true));
```
slozier: By calling the worker directly it looks like `RemoveFrameList` is no longer called (I have no idea what it does).

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -2422,7 +2422,7 @@ public static partial class PythonOps {
             return e;
         }
 
-        private static Exception MakeExceptionWorker(CodeContext/*!*/ context, object type, object value, object traceback, bool forRethrow) {
+        internal static Exception MakeExceptionWorker(CodeContext/*!*/ context, object type, object value, object traceback, bool forRethrow, bool forGenerator=false) {
```
slozier: I think this is going to lead to some intellisense confusion (do I use MakeException or MakeExceptionWorker). Instead of exposing `MakeExceptionWorker` internally could we maybe add an (internal) `MakeExceptionForGeneratorTypeError` or something of the sort?

PR #174, opened 2017-06-21
Fix #4

### LINKED ISSUE #4: ssl.read tries to allocate a 2GB MemoryStream if no len parameter is passed

slozier:
Looks like the `len` argument is not actually optional even though the doc string seems to say it is.


slide:
This looks good, we are missing the ability to pass in a buffer as an optional parameter that is used for the data.

See https://github.com/python/cpython/blob/2.7/Modules/_ssl.c#L1790

I'll file an issue so that we implement that later.

PR #181, opened 2017-06-22
Fix #16

### LINKED ISSUE #10: Message for trying to raise 'str' inconsistent with CPython

### LINKED ISSUE #16: Error handler for non-builtin encodings are not being called
_From @slozier on May 11, 2017 19:8_

Here's the repro code:
```Python
def test_unicode_error():
    # test with a standard error handler
    res = u"\xac\u1234\u20ac\u8000".encode("rot_13", "backslashreplace")
    assert res == "\xac\\h1234\\h20np\\h8000"

    # test with a custom error handler
    from codecs import register_error
    def handler(ex):
        return (u"", ex.end)
    register_error("test_unicode_error", handler)
    res = u"\xac\u1234\u20ac\u8000".encode("rot_13", "test_unicode_error")
    assert res == "\xac"

test_unicode_error()
```

_Copied from original issue: IronLanguages/main#1639_

slide:
The commit that says "Fix #10" is wrong...

This fixes the charmap encoding stuff

### DIFF
--- a/Src/IronPython.Modules/_codecs.cs
+++ b/Src/IronPython.Modules/_codecs.cs
@@ -294,13 +294,8 @@ namespace IronPython.Modules {
             );
         }
 
-        public static object readbuffer_encode(string input, string errors=null) {
+        public static object readbuffer_encode([BytesConversion]string input, string errors=null) {
             return PythonTuple.MakeTuple(input, input.Length);
-        }
-
-        public static object readbuffer_encode(IPythonBufferable buffer, string errors=null) {
-            string val = buffer.ToString();
-            return PythonTuple.MakeTuple(val, val.Length);
         }
 
         public static void register(CodeContext/*!*/ context, object search_function) {
--- a/Src/IronPython.Modules/_codecs.cs
+++ b/Src/IronPython.Modules/_codecs.cs
@@ -824,7 +824,7 @@ namespace IronPython.Modules {
             foreach(var k in _map) {
                 if(k.Key is string) {
                     var s = (string)k.Key;
-                    if(s.Length > 0)
+                    if(s.Length == 1)
                         _map[(int)s[0]] = k.Value;
                 }
             }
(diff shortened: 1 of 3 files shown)


Src/IronPython.Modules/_codecs.cs
```diff
@@ -366,8 +294,13 @@ public class EncodingMap {
             );
         }
 
-        public static object readbuffer_encode(string input) {
-            throw PythonOps.NotImplementedError("readbuffer_encode");
+        public static object readbuffer_encode(string input, string errors=null) {
+            return PythonTuple.MakeTuple(input, input.Length);
+        }
+
+        public static object readbuffer_encode(IPythonBufferable buffer, string errors=null) {
```
slozier: Seems like an odd overload, Are you sure `IPythonBufferable` is the type you want?

slide: No, that shouldn't have been part of the push

Src/IronPython.Modules/_codecs.cs
```diff
@@ -68,146 +72,70 @@ public static class PythonCodecs {
...
-                    if (errors == "strict") {
-                        throw PythonOps.UnicodeDecodeError("failed to find key in mapping");
-                    }
-                    res.Append("\ufffd");
-                } else {
-                    res.Append(map[input[i]]);
-                }
-            }
-            return PythonTuple.MakeTuple(res.ToString(), res.Length);
+        public static object charmap_encode(CodeContext context, [BytesConversion]string input, string errors = "strict", IDictionary<object, object> map = null) {
```
slozier: Return `PythonTuple` instead of `object`?

slide: I don't think the python side cares, but if you think its a good idea, I'll do it

PR #184, opened 2017-06-24
Fix #23

### LINKED ISSUE #23: Deprecation warning no longer raised on object.__init__
_From @slozier on March 23, 2017 14:44_

While IronLanguages/main@042e870 resolved issue #1526 it incorrectly removed deprecation warnings for the following type of calls:
```
object.__init__(object, None)
```

_Copied from original issue: IronLanguages/main#1614_

slide:
Re-add code that was removed and fix ref type in _weakref

### DIFF
--- a/Src/IronPython.Modules/_weakref.cs
+++ b/Src/IronPython.Modules/_weakref.cs
@@ -120,7 +120,7 @@ namespace IronPython.Modules {
             #region Constructors            
 
             public @ref(CodeContext context, object @object, object callback=null) {
-                __init__(context, @object, callback);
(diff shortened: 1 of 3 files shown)


slide:
We could run something with -Qwarnall and trap the warnings using the warnings trapping in test_support

slozier:
Yeah, looks like the `assertWarns` in `IronPythonTestCase` should do the trick.

slide:
Do you just want 

```python
object.__init__(object, None)
```

for now, or where there other warnings you wanted to check?

Issue #200, opened 2017-07-09
Remove unnecessary features


slozier:
I did a quick survey of the feature tags and identified the ones which are widely supported (net40+ and netstandard1.1+). I propose we remove the conditional code for these features:
- FEATURE_APTCA
- FEATURE_COMPRESSION (this one appears limited to GZipStream)
- FEATURE_CUSTOM_MODIFIERS
- FEATURE_DEFAULT_PARAMETER_VALUE
- FEATURE_ENUM_NAMES_VALUES
- FEATURE_GETINVALIDFILENAMECHARS
- FEATURE_NUMERICS
- FEATURE_STOPWATCH
- FEATURE_TASKS

slide:
Sounds great

PR #202, opened 2017-07-10
Remove unnecessary features

slozier:
The rest of IronLanguages/ironpython2#200.


Build/net45.props
```diff
@@ -10,14 +10,11 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <Features>FEATURE_CUSTOM_MODIFIERS</Features>
     <Features>$(Features);FEATURE_NATIVE</Features>
```
slide: Need to remove the variable and semicolon

slozier: I thought about that, but then I saw that net40.props didn't do it so I left it as is. If you think it'll cause issues I can change it.

PR #214, opened 2017-07-15
Update to VS2017

slozier:
Similar to IronLanguages/ironpython3#323


Src/IronPythonTest/IronPythonTest.csproj
```diff
@@ -1,105 +1,42 @@
...
     <EmbeddedResource Include="Cases\IronPythonCasesManifest.ini" />
-  </ItemGroup>
-  <ItemGroup>
     <EmbeddedResource Include="Cases\AllCPythonCasesManifest.ini" />
+    <EmbeddedResource Include="Cases\StandardCPythonCasesManifest.ini" />
   </ItemGroup>
   <ItemGroup>
-    <EmbeddedResource Include="Cases\StandardCPythonCasesManifest.ini" />
+    <None Include="..\App.Debug.config" Link="App.Debug.config" />
+    <None Include="..\App.Release.config" Link="App.Release.config" />
```
gpetrou: I think these two are not needed?

Src/IronPython.Modules/fcntl.cs
```diff
@@ -1,8 +0,0 @@
-
```
slide: Why remove this file?

slozier:
Some stuff I noted while working on this (maybe file issues):
- The app.config files aren't getting deployed. In any case they are out of date (version numbers and keys).
- We used to have two signing keys, one for release and one for debug, this appears to no longer be the case. I'm fine with this however it makes it harder to differentiate between a release and a debug DLL. Is there some other identifier?
- The staging/packaging is probably broken (or will be with this PR).

slide:
Yes, I've been waiting for something like this before updating the packaging stuff.

Src/IronPython.Modules/fcntl.cs
slozier: It wasn't included in the project before (and didn't compile). Including it also causes failures in stuff that was working before (e.g. tempfile).

slozier:
@slide Any idea for the Travis failures? I'm not sure if it's the project reorg or the DLR update which is causing them.

slozier:
Close and reopen to test.

slozier:
Close and reopen to test.

PR #217, opened 2017-07-18
Use FindRoot instead of fixed path

slozier:


Src/IronPythonTest/Cases/CaseExecuter.cs
```diff
@@ -37,7 +37,7 @@ class CaseExecuter {
             return engine;
         }
 
-        private static string FindRoot() {
+        internal static string FindRoot() {
```
slide: Maybe this should be moved to a more common location?

slozier: Any suggestions?

slide: Not really...I guess this is as good as anything for now.

PR #238, opened 2017-09-07
Fix #218

### LINKED ISSUE #218: Switch IronPythonTest from an exe to a dll
If possible we should change IronPythonTest to generate an DLL which we can run using the nunit3.consolerunner. This would allow us to also run the tests in 32-bit on a 64-bit system.

slide:
- Migrate IronPythonTest to dll.
- Cleanup some files


Src/StdLib/MakeModuleList.py
```diff
@@ -17,7 +17,7 @@
 # Usage: ipy getModuleList.py <output directory>
 
 #--Imports---------------------------------------------------------------------
-import sys, nt, os
+import sys, os
```
slozier: Do we still need this file? Couldn't we just use a static pyproj file with a glob pattern? Maybe something to look into for a separate PR...

Src/App.Release.config
```diff
@@ -1,10 +1,6 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <!-- Release version of App.config -->
 <configuration>
-  <configSections>
-    <section name="microsoft.scripting" type="Microsoft.Scripting.Hosting.Configuration.Section, Microsoft.Scripting, Version=1.1.2.22, Culture=neutral, PublicKeyToken=7f709c5b713576e1" requirePermission="false"/>
```
slozier: Any reason this is completely removed for release but included in debug?

Src/StdLib/MakeModuleList.py
slide: Yes, we actually do. This file checks for modules that don't work with IronPython and does not include them in the release area.

Src/App.Release.config
slide: I forgot to remove it from the debug version.

PR #244, opened 2017-10-03
Re-add test (mono was fixed)

slide:
Re-enable test now that Mono is fixed


slide:
Meh, I messed up my local tree...

PR #247, opened 2017-10-26
 Remove condition on TargetFrameworks and update NuGet packages

slozier:
Removing the condition prevents VS2017 from trying to trigger a project migration.


make.cmd
```diff
@@ -90,64 +92,65 @@ echo No target 'test'. Try 'test-smoke', 'test-ironpython', 'test-cpython', or '
 goto :exit
 
 :test-smoke
-for %%f in ("%FRAMEWORKS:,=" "%") do (
+ for %%f in ("%FRAMEWORKS:,=" "%") do (
```
slide: Space?

PR #259, opened 2017-11-07
Use powershell for build

slide:


make.ps1
```diff
@@ -0,0 +1,138 @@
...
+
+$_BASEDIR = Split-Path -Path $MyInvocation.MyCommand.Definition -Parent
+
+$_VSWHERE = [System.IO.Path]::Combine(${env:ProgramFiles(x86)}, 'Microsoft Visual Studio\Installer\vswhere.exe')
+$_VSINSTPATH = ''
+
+if([System.IO.File]::Exists($_VSWHERE)) {
+    $_VSINSTPATH = & "$_VSWHERE" -latest -requires Microsoft.Component.MSBuild -property installationPath
+} else {
+    Write-Error "Visual Studio 2017 15.2 or later is required"
```
gpetrou: Should this be 15.3?

make.ps1
```diff
@@ -0,0 +1,138 @@
...
+            "all" = "";
+            "smoke" = "TestCategory=StandardCPython";
+            "cpython" = "TestCategory=StandardCPython | TestCategory=AllCPython";
+            "ironpython" = "TestCategory=IronPython";
+        }
+    }
+}
+
+function Main([String] $target, [String] $configuration) {
+    msbuild Build.proj /t:$target /p:BuildFlavour=$configuration /verbosity:minimal /nologo /p:Platform="Any CPU" /bl:build-$target.binlog
```
gpetrou: Should this have /m parameter as well?

make.ps1
```diff
@@ -0,0 +1,138 @@
+[CmdletBinding()]
```
gpetrou: Could this be modified for Travis usage as well?

make.ps1
slide: It historically has not, but we could add it

make.ps1
slozier: While we may required 15.3 to build for .NET Core the vswhere check will pass with 15.2 or later.

make.ps1
slozier: Does our build process work properly in parallel? If so then might as well add it.

make.ps1
```diff
@@ -0,0 +1,138 @@
...
+            "__FILTER__" = $filter;
+            "__BASEDIR__" = $_BASEDIR;
+        };
+
+        $runner = $_FRAMEWORKS[$framework]["runner"]
+        # make a copy of the args array
+        $_TempCliXMLString = [System.Management.Automation.PSSerializer]::Serialize($_FRAMEWORKS[$framework]["args"], [int32]::MaxValue)
+        [Object[]] $args = [System.Management.Automation.PSSerializer]::Deserialize($_TempCliXMLString)
+        # replace the placeholders with actual values
+        for([int32] $i = 0; $i -lt $args.Length; $i++) {
```
slozier: Powershell is case insensitive in its types or is int32 valid? I would have expected `int` or `Int32`

make.ps1
```diff
@@ -0,0 +1,138 @@
...
+    "ngen"          {
+        $imagePath = [System.IO.Path]::Combine($_BASEDIR, "bin\$configuration\net45\ipy.exe")
+        & "${env:SystemRoot}\Microsoft.NET\Framework\v4.0.30319\ngen.exe" install $imagePath 
+    }
+
+    "debug"         { Main "Build" "Debug" }
+    "clean-debug"   { Main "Clean" "Debug" }
+    "stage-debug"   { Main "Stage" "Debug" }
+    "package-debug" { Main "Package" "Debug" }
+    "test-debug-*"  { Test $target.Substring(10) "Debug" $frameworks }
```
slozier: Is it smart enough to figure out that test-debug-* isn't contained in test-*?

Src/IronPythonTest/IronPythonTest.csproj
```diff
@@ -20,7 +20,7 @@
     <PackageReference Include="NUnit.ConsoleRunner" Version="3.7.0" />
     <PackageReference Include="NUnitLite" Version="3.8.1" />
   </ItemGroup>
-  <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp2.0' ">
+  <ItemGroup Condition=" '$(TargetFramework)' != 'net40' ">
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
     <PackageReference Include="NUnit3TestAdapter" Version="3.8.0" />
     <PackageReference Include="Mono.Cecil" Version="0.10.0-beta6" />
```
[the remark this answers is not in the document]
slide: It's required for the NUnit3TestAdapter

make.ps1
slide: No, good catch.

make.ps1
slide: `int32` is valid. I'll change to `int`.

make.ps1
slide: I built locally and it didn't die, so I'll add it

Issue #266, opened 2017-11-21
WeakRefTracker class is not thread safe


tinyg:
Hi All,

We have experienced a concurrency issue with OrderDicts which causes various errors including: "Index out of range", "weakref has gone away" and "array copy" exceptions.

This can be reproduced by creating lots of OrderDict instances in different threads at the same time. OrderDict uses the abc module, which in turn uses WeakSet to track metadata, which in turn uses weakrefs which seem to be the root cause.

Attached is a script which reproduces the issue (perhaps 1 in every 5 runs will fail). eg, in our local release version:

```
Exception in thread Thread-8:
Traceback (most recent call last):
  File "C:\TFS\SDKS\ironpython\2.7.7\Lib\threading.py", line 801, in _Thread__bo
otstrap_inner
    self.run()
  File "C:\TFS\SDKS\ironpython\2.7.7\Lib\threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "ironpython_weakref.py", line 37, in create_stuff
    [make_list() for _ in range(10)]
  File "ironpython_weakref.py", line 29, in make_list
    return collections.OrderedDict(
  File "C:\TFS\SDKS\ironpython\2.7.7\Lib\collections.py", line 69, in __init__
    self.__update(*args, **kwds)
  File "C:\TFS\SDKS\ironpython\2.7.7\Lib\_abcoll.py", line 571, in update
    for key, value in other:
  File "ironpython_weakref.py", line 31, in <genexpr>
    for __ in range(8)
  File "C:\TFS\SDKS\ironpython\2.7.7\Lib\collections.py", line 69, in __init__
    self.__update(*args, **kwds)
  File "C:\TFS\SDKS\ironpython\2.7.7\Lib\abc.py", line 151, in __subclasscheck__

    if subclass in cls._abc_cache:
  File "C:\TFS\SDKS\ironpython\2.7.7\Lib\_weakrefset.py", line 72, in __contains
__
    wr = ref(item)
IndexError: Index was outside the bounds of the array.
```

and

```
Exception in thread Thread-15:
Traceback (most recent call last):
  File "C:\TFS\SDKS\ironpython\2.7.7\Lib\threading.py", line 801, in _Thread__bo
otstrap_inner
    self.run()
  File "C:\TFS\SDKS\ironpython\2.7.7\Lib\threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
[...]

PR #269, opened 2017-11-21
Thread safety for weakrefs (#266)

tinyg:
Proposed fix for #266 

Wrap all access to _callbacks in a ReaderWriterLockSlim to prevent concurrency issues in the list.

### DIFF
--- a/Src/IronPython/Runtime/WeakRef.cs
+++ b/Src/IronPython/Runtime/WeakRef.cs
@@ -26,6 +26,8 @@ using IronPython.Runtime.Binding;
 using IronPython.Runtime.Operations;
 
 namespace IronPython.Runtime {
+    using System.Threading;
+
     /// <summary>
     /// single finalizable instance used to track and deliver all the 
     /// callbacks for a single object that has been weakly referenced by
@@ -74,6 +76,7 @@ namespace IronPython.Runtime {
         }
 
         private readonly long _targetId;
+        private readonly ReaderWriterLockSlim _lock = new ReaderWriterLockSlim();
         private readonly List<CallbackInfo> _callbacks = new List<CallbackInfo>(1);
 
         public WeakRefTracker(IWeakReferenceable target) {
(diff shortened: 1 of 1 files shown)


Src/IronPython/Runtime/WeakRef.cs
```diff
@@ -26,6 +26,8 @@
 using IronPython.Runtime.Operations;
 
 namespace IronPython.Runtime {
+    using System.Threading;
```
slide: Can you move this near the other usings?

Src/IronPython/Runtime/WeakRef.cs
```diff
@@ -91,67 +94,133 @@ struct CallbackInfo {
         }
 
         public void ChainCallback(object callback, object weakRef) {
-            _callbacks.Add(new CallbackInfo(callback, weakRef));
+            _lock.EnterWriteLock();
```
slide: Can we not lock on the _callbacks directly?

tinyg: locking on callbacks would be slightly less efficient - the read write lock allows simultaneous read access but synchronized write access whereas a simple lock will synchronize on all access...

Src/IronPython/Runtime/WeakRef.cs
tinyg: sorry - yes this was our local stylecop policy getting over excited.

tinyg:
mm  - I've let my test script run for longer (ie 100 runs now instead of 30) and I can see we still get the TypeError: Weakref has gone away! issue perhaps 3/100 runs.

Looks like _weakrefset.py in the stdlib has:

```
    def __contains__(self, item):
        try:
            wr = ref(item)
        except TypeError:
            return False
        return wr in self.data
```

however the weakref could be disposed between creating it and returning wr in self.data. The following change fixes it (no errors in 1000 runs):

```
    def __contains__(self, item):
        try:
            wr = ref(item)
            return wr in self.data
        except TypeError:
            return False
```

How do you manage changes to the standard library code? Can I just submit another changeset with this fix in Src\StdLib\Lib\_weakrefset.py ?

tinyg:
OK - I've added the extra fix - I can roll it back if there's a different way to do it.

Src/StdLib/Lib/_weakrefset.py
```diff
@@ -70,9 +70,9 @@ def __len__(self):
     def __contains__(self, item):
         try:
             wr = ref(item)
+            return wr in self.data
```
slide: Changes to the stdlib need to be reviewed quite a bit, we want to stay as close to the CPython stdlib as possible. In some cases, it's not possible, but we want to make sure that is the case before incorporating changes.

Src/IronPython/Runtime/WeakRef.cs
```diff
@@ -91,67 +94,133 @@ struct CallbackInfo {
...
+            try
```
slide: Can you reformat so that the braces are on the same lines as the try and finally? This is the style we use in the code.

Src/IronPython/Runtime/WeakRef.cs
```diff
@@ -91,67 +94,133 @@ struct CallbackInfo {
...
-                            // Going through PythonCalls / Rules requires the types be public.
-                            // Explicit check so that we can keep InstanceFinalizer internal.
-                            fin.CallDirect(DefaultContext.Default);
-                                
-                        } else {
-                            // Non-instance finalizer goes through normal call mechanism.
-                            object weakRef = ci.WeakRef;
-                            if (weakRef != null)
-                                PythonCalls.Call(ci.Callback, weakRef);
+            _lock.EnterWriteLock();
```
slozier: Is there a reason why this one can't be a read lock?

Src/StdLib/Lib/_weakrefset.py
slozier: I'm not saying there's no issue here, but I'm not really sure I follow the reasoning for this change. What exactly is getting GCed here and how does this fix the issue? Looking at the `abc` module it seems to be maintaining the reference to `item` so it shouldn't be collected. And the weakref object certainly isn't getting collected before the end of the operations.

tinyg: I completely understand. I’m surprised that line isn’t an issue in CPython - perhaps cpython couldnt context switch between the two lines? I don’t really understand cpython’s thread model enough to comment. From .nets perspective there’s definitely a potential race condition there tho

I guess the new structure is functionally the same as the old in this case so I could potentially submit a CPython pull request also (probably get rejected but you never know)

Src/IronPython/Runtime/WeakRef.cs
tinyg: I guess its intent rather than actual action. This block is trying run all the callbacks and then remove them as far as I can see - while it doesn't actually edit the list, there's an implicit _callbacks.Clear() since this is a finaliser. My personal style would be to be conservative and assume this code is "entitled" to edit the list and therefore protect it as if that were the case (and possibly prevent a future bug if someone does edit the list) . Happy to change it if you prefer to leave it as a read lock.

It probably doesn't matter either way since the finaliser should be called when there are no references to the WeakRefTracker. Possibly it doesn't need a lock at all...

Src/StdLib/Lib/_weakrefset.py
tinyg: Yes - it does seem odd that having the item as a parameter is not forcing that item in memory.

I'm trying to think of other possible ways you could get the error from that line...

What about if one of the other items in the set has been freed - perhaps CPython is just testing the weakref provided against the pointer location of the weakrefs in the set to establish equality whereas IronPython might be dereferencing each item in the set and comparing the inner items? That would cause the symptom I'm seeing... I guess it would mean the fix should be in the equality check tho rather than what I proposed.

I'll see if I can debug into the comparison and see what its doing

tinyg: So I spent today playing with this and, while I dont really understand why, my original diagnosis appears to be correct.

The exception comes from the __hash__ function of weakref.ref at _weakref.cs line 219 when the "in" operator is called. So either the items in the set have been freed and their hashes fail or the item in wr has been freed and its hash fails.

You can run the following code to explore the behaviour of the weakref.ref hash function and get the exception:

```
class A(): pass

# first show that weakref.ref hash works normally
a = A()
import weakref
r = weakref.ref(a)
print hash(r)
# prints hash

# now show that the hash is cached
del a
import gc
gc.collect(2)
print hash(r)
# prints same hash

# now show that, if you dont calculate the hash before you dispose the item then it fails
del r
a = A()
r = weakref.ref(a)
del a
gc.collect(2)
print hash(r)
# error
```

So, looking at the implementation of Set, you can see that it caches the hashes as soon as you add an item to the list. This suggests that its "item" that has been freed, not other items in the set.

To confirm this, you can undo my original fix and instead force the cache of the hash:

```
    def __contains__(self, item):
        try:
            wr = ref(item)
            hash(wr)
        except TypeError:
            return False
        return wr in self.data
```

Running the original test 100 times shows this also fixes the issue.

From other testing I've done, I've seen that try/except blocks seem to sometimes (but not reliably?) form scope boundaries for IronPython so my suspicion is that item is being "released" when leaving the try/except block and in the 3 out of 100 case the GC fires just at the wrong time and blows item away (I'm not sure why, as you say, you'd expect the stack frame entering __contains__ to be holding a reference to item). I cant reproduce this by manually calling the gc so I feel I'm on shaky ground with this assertion. 

Have a look at the attached testscope script for various tests around scoping and gc - there's some oddities and I did manage to sort of recreate the bug with slightly modified code where "item" is itself a weakref...
[testscope.zip](https://github.com/IronLanguages/ironpython2/files/1498995/testscope.zip)

Src/IronPython/Runtime/WeakRef.cs
tinyg: No probs - I'll put in another commit for this

tinyg: ok - I've updated the styles - can you check that it looks ok now?

Src/StdLib/Lib/_weakrefset.py
tinyg: BTW - I also ran the original bug script against CPython and cannot reproduce this TypeError there even tho I can see the ipy and python code are pretty much line-for-line ports all the way down. Some aspect of CPython's threading / GC must be happily blocking this issue from occurring there.

slozier: Thanks for your analysis. I concur that the only way the exception would occur is if `item` has been GCed before the hashing occurs. I just don't get how it happens since I would expect `item` to be held alive until the end of the function call (as your `test3` shows).

I'm fine with the proposed change to `_weakrefset.py `, however you should mark it up with a comment referencing the issue or PR otherwise it may get reverted when we update the standard library to sync up with CPython.

As for the testscope script here are my comments:
- `test3` is behaving as expected in that argument should be held alive until the end of the function call.
- `test4` only being cleaned up in `test5` is certainly unexpected - seems like a bug
- `test6` is also unexpected and seems like a bug
- `test7` is behaving as expected, with IPy there should be two possible outcomes:
  - GC occurs between the assignment to `a` and the call to `test7` in which case the value of `b` is `None` which gives the same result as CPython
  - Otherwise the GC hasn't occurred and the value of `b` is `A(7)`. In this case `b` should be held alive until the function returns
- `test8` is also behaving as expected. Like test7 there are two possible outcomes. The value of `b` is always the weakref but its target has either been collected of not:
  - GC occurs between the assignment to `a` and the call to `test8` in which case the target of `b` is `None` and we get the same result as CPython.
  - Otherwise the GC hasn't occurred and the target of `b` is still `A(8)`. However in this case since `b` is a weakref to `A(8)` the GC can collect still clean up `A(8)`.

tinyg: Added one more commit to comment the change in _weakrefset.py

Thanks for looking at testscope - I agree with your comments. Should I add issues to cover those test cases?

slozier: Sure, better off adding an issue to cover those cases otherwise I'm likely to forget about it.

tinyg:
Fixed two further formatting issues - hopefully all done now

PR #270, opened 2017-11-23
Cleanup and test updates

slide:


Tests/modules/io_related/test_cStringIO.py
```diff
@@ -20,6 +20,8 @@
 import cStringIO
 import unittest
 
+from iptest import rnu_test
```
slozier: Typo: rnu_test

slide: Thanks, I thought I had run everything prior, but maybe made a last edit before committing

PR #277, opened 2017-11-28
Update StdLib to 2.7.14

slozier:
- Updated using the StdLib from a clean 2.7.14 windows installation.
- Removed idlelib and lib-tk since we don't support them and all we had were the tests.
- Did not update ensurepip.
- Split it in three commits just to make it easier to review:
  - First commit applies all the stdlib changes and overrides any cli modifications.
  - Second commit applies the cli modifications.
  - Fixes/disables failing tests


Src/StdLib/Lib/subprocess.py
```diff
@@ -1510,13 +1190,16 @@ def _dup2(a, b):
 
         def _handle_exitstatus(self, sts, _WIFSIGNALED=os.WIFSIGNALED,
                 _WTERMSIG=os.WTERMSIG, _WIFEXITED=os.WIFEXITED,
-                _WEXITSTATUS=os.WEXITSTATUS):
+                _WEXITSTATUS=os.WEXITSTATUS, _WIFSTOPPED=os.WIFSTOPPED,
```
slide: Do we have all these properties on the os module?

slozier: No, but they're in an unsupported codepath right now (if not windows and not posix).

PR #282, opened 2017-11-30
Add clrtype and test

slide:


Src/StdLib/Lib/clrtype.py
```diff
@@ -223,13 +227,7 @@ def define_interface(typename, bases):
...
             name = AssemblyName("interfaces")
             access = AssemblyBuilderAccess.Run
-            # Workaround for http://ironpython.codeplex.com/WorkItem/View.aspx?WorkItemId=25330
-            for i in xrange(1000):
-                try:
-                    ad = AppDomain.CurrentDomain
-                    assembly_builder = ad.DefineDynamicAssembly(name, access)
-                except: pass
-                else: break
+            assembly_builder = AssemblyBuilder.DefineDynamicAssembly(name, access)
```
slozier: Did you test this with .NET 4.0? I don't think `AssemblyBuilder.DefineDynamicAssembly` is available. You could try using `Microsoft.Scripting.Utils.ReflectionUtils.DefineDynamicAssembly` or `IronPython.Runtime.Operations.PythonOps.DefineDynamicAssembly` instead (don't ask me why we have two ways of doing the same thing...)

slide: No, I didn't. I'm not too worried about .NET 4.0 at this point.

PR #283, opened 2017-12-03
Include XML files in the final package

gpetrou:


slide:
This doesn't quite work correctly yet. There is a DLLs directory created in the nuget package that shouldn't be there. Also, for the zip package, we should probably include the XML files for the DLR stuff.

PR #286, opened 2017-12-05
Fix SSL certificate parsing issues

slozier:
Resolves a few issues with certificate parsing. Fixes https://github.com/IronLanguages/ironpython2/issues/57.


slide:
Is there a test that we can enable that would catch this?

slozier:
I don't know how we can test `subjectAltName` since it depends in the underlying OS locale. The other stuff we should be able to test if you think it's worth it. The whole `_ssl` module needs work.

PR #287, opened 2017-12-06
Fix #272

### LINKED ISSUE #272: test_importpkg doesn't clean up temporary files
`test_importpkg` creates the `ImportTestDir` and doesn't remove it after it's done.

slide:
Cleanup generated directories

### DIFF
--- a/Src/IronPython/Lib/iptest/file_util.py
+++ b/Src/IronPython/Lib/iptest/file_util.py
@@ -75,7 +75,7 @@ class FileUtil(object):
             try:    os.remove(f)
             except: pass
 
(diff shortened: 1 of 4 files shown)


Src/IronPython/Lib/iptest/file_util.py
```diff
@@ -76,155 +76,20 @@ def delete_files(self, *files):
             except: pass
 
     def clean_directory(self, path, remove=False):
+        import shutil
```
slozier: I'm tempted to say we should just remove the `remove` argument (always remove the directory).

PR #290, opened 2017-12-06
Performance improvements for _struct

slozier:


slozier:
I just did some simple tests with timeit:
```Python
timeit.timeit("unpack(s)", setup="import struct\nunpack = struct.Struct('<H').unpack\ns='\x12\x34'")
```
On my machine it's about 2.5-3x faster. Assuming the bulk of the memory usage is in the array then it's 1/3 the memory usage (since it gets rid of two copies of the data in List<object>).

PR #291, opened 2017-12-06
Fix #288

### LINKED ISSUE #288: IronPython2 raises error when C Python does not
In test_ensurepip, there is a mocking library used. In one function (below) IronPython2 raises an error:

`SyntaxError: unqualified exec is not allowed in function '_set_signature' because it contains a nested function with free variables`

Where C Python does not.

```python
def _set_signature(mock, original, instance=False):
    # creates a function with signature (*args, **kwargs) that delegates to a
    # mock. It still does signature checking by calling a lambda with the same
    # signature as the original.
    if not _callable(original):
        return

    skipfirst = isinstance(original, type)
    result = _get_signature_object(original, instance, skipfirst)
    if result is None:
        return
    func, sig = result
    def checksig(*args, **kwargs):
        sig.bind(*args, **kwargs)
    _copy_func_details(func, checksig)

    name = original.__name__
    if not name.isidentifier():
        name = 'funcopy'
    context = {'_checksig_': checksig, 'mock': mock}
    src = """def %s(*args, **kwargs):
    _checksig_(*args, **kwargs)
    return mock(*args, **kwargs)""" % name
    exec (src, context)
    funcopy = context[name]
    _setup_func(funcopy, mock)
    return funcopy
```

slide:
- Fix parser for exec function call form
- Update test_ensurepip.py to work on IronPython

### DIFF
--- a/Src/IronPython/Compiler/Parser.cs
+++ b/Src/IronPython/Compiler/Parser.cs
@@ -951,24 +951,23 @@ namespace IronPython.Compiler {
             Eat(TokenKind.KeywordExec);
             var start = GetStart();
             Expression code, locals = null, globals = null;
-            if (MaybeEat(TokenKind.LeftParenthesis)) { // 'exec' '(' expr ',' expression [',' expression] ')'
-                code = ParseExpr();
+
+            code = ParseExpr();
+            if (code is TupleExpression t) {
+                if (t.Items.Count == 2 || t.Items.Count == 3) {
+                    globals = t.Items[1];
+                    if (t.Items.Count == 3) {
+                        locals = t.Items[2];
+                    }
+                    code = t.Items[0];
+                }
+            } else if (MaybeEat(TokenKind.KeywordIn)) { // 'exec' expr 'in' expression [',' expression]
+                globals = ParseExpression();
                 if (MaybeEat(TokenKind.Comma)) {
-                    globals = ParseExpression();
-                    if (MaybeEat(TokenKind.Comma)) {
-                        locals = ParseExpression();
-                    }
+                    locals = ParseExpression();
                 }
-                Eat(TokenKind.RightParenthesis);
-            } else {
-                code = ParseExpr();
-                if (MaybeEat(TokenKind.KeywordIn)) { // 'exec' expr 'in' expression [',' expression]
-                    globals = ParseExpression();
-                    if (MaybeEat(TokenKind.Comma)) {
-                        locals = ParseExpression();
-                    }
-                }
-            }
+            }            
+
             ExecStatement ret = new ExecStatement(code, locals, globals);
             ret.SetLoc(_globalParent, start, GetEnd());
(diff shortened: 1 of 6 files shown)


slozier:
This will raise `SyntaxError` instead of a `TypeError` for things like:
```Python
exec("",{},{},1)
```
Something like:
```C#
if (MaybeEat(TokenKind.KeywordIn)) { // 'exec' expr 'in' expression [',' expression]
    globals = ParseExpression();
    if (MaybeEat(TokenKind.Comma)) {
        locals = ParseExpression();
    }
}
if (globals == null && code is TupleExpression tupleExpr) {
    if (tupleExpr.Items.Count == 2) {
        code = tupleExpr.Items[0];
        globals = tupleExpr.Items[1];
    } else if (tupleExpr.Items.Count == 3) {
        code = tupleExpr.Items[0];
        globals = tupleExpr.Items[1];
        locals = tupleExpr.Items[2];
    }
}
```
is probably more technically correct.

Also, we should probably have some tests specifically for this (other than test_ensurepip).

Src/IronPython/Compiler/Parser.cs
```diff
@@ -951,13 +951,23 @@ public class Parser : IDisposable { // TODO: remove IDisposable
             Eat(TokenKind.KeywordExec);
             var start = GetStart();
             Expression code, locals = null, globals = null;
+
             code = ParseExpr();
-            if (MaybeEat(TokenKind.KeywordIn)) {
+            if (code is TupleExpression t) {
```
slozier: I'm pretty sure this needs to be done the other way around (like in my example) otherwise you get a `SyntaxError` instead of a `TypeError`:
```Python
exec ("",{}) in {}
```

Src/StdLib/Lib/test/test_ensurepip.py
```diff
@@ -38,11 +38,19 @@ class TestBootstrap(EnsurepipMixin, unittest.TestCase):
     def test_basic_bootstrapping(self):
         ensurepip.bootstrap()
 
+        if sys.platform == 'cli':
+            args = [
+                    "install", "--no-index", "--find-links",
+                    mock.ANY, "--no-compile", "setuptools", "pip",
```
slozier: So we're just adding no-compile everywhere for the cli? Does the args call work with an empty string? Maybe we could be slightly less verbose:
```
noCompileIfCli = "--no-compile" is sys.platform == "cli" else "" # this could be a global
[
    "install", "--no-index", "--find-links",
    mock.ANY, noCompileIfCli, "setuptools", "pip",
]
```
Anyway, just a thought...

slide: This doesn't work on C python, it fails every test because of the difference in argument lists.

slozier: Alright, we can keep it as is.

PR #295, opened 2017-12-07
Fix exception printing (remove extra newline)

slide:
Our exception dump had an extra newline between the source line and caret pointing to the issue


slozier:
There's a slightly different implementation of this in [ipy3](https://github.com/IronLanguages/ironpython3/pull/236/files#diff-ccf15637539eebe2e10c17b4dc36237e). Would be nice to be consistent. I have no preference either way.

While you're in there, maybe remove the extra line endings at the end?
IronPython:
```
>>> raise Exception
  File "<stdin>", line 1, in <module>
Exception

>>>
```

CPython:
```
>>> raise Exception
  File "<stdin>", line 1, in <module>
Exception
>>>
```

slide:
I'll make them consistent and fix the other newline as well.

Issue #292, opened 2017-12-07
test_grammar failures


slide:
We currently fail the following in test_grammar.py

- [x]  testPlainIntegers - `self.assertRaises(SyntaxError, eval, "0x")` - should raise SyntaxError, does not
- [ ]  testAssert2 - `self.assertEqual(len(e.args), 0)` - length of args is 1 instead of 0
- [x]  test_float_exponent_tokenization - `self.assertEqual(1 if 1else 0, 1)` - we don't parse this right now

slide:
testPlainIntegers is fixed via #294

PR #297, opened 2017-12-11
Fix assertion error grammar

slozier:
Fixes https://github.com/IronLanguages/ironpython2/issues/292.

### DIFF
--- a/Src/IronPython/Runtime/Operations/PythonOps.cs
+++ b/Src/IronPython/Runtime/Operations/PythonOps.cs
@@ -1523,7 +1523,7 @@ namespace IronPython.Runtime.Operations {
                 throw PythonOps.CreateThrowable(DynamicHelpers.GetPythonType(obj));
             }
 
-            throw PythonOps.AssertionError("{0}");
+            throw PythonOps.AssertionError();
         }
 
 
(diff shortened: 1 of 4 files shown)


Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -1513,27 +1513,35 @@ public static partial class PythonOps {
...
+                throw PythonOps.CreateThrowable(DynamicHelpers.GetPythonType(obj));
             }
 
-            if (builtins._storage.TryGetValue("AssertionError", out obj)) {
-                PythonType type = obj as PythonType;
-                if (type != null) {
-                    throw PythonOps.CreateThrowable(type, message);
-                } else {
-                    throw PythonOps.CreateThrowable(DynamicHelpers.GetPythonType(obj), message);
+            throw PythonOps.AssertionError("{0}");
```
slide: Just for my benefit, can you add a comment on why we call this with a format specifier?

slozier: Because I just mindlessly removed the message attribute. There shouldn't be an argument there... Will fix tomorrow.

slide:
Close and reopen for build

Src/IronPython/Runtime/Operations/PythonOps.cs
```diff
@@ -1523,7 +1523,7 @@ public static partial class PythonOps {
                 throw PythonOps.CreateThrowable(DynamicHelpers.GetPythonType(obj));
             }
 
-            throw PythonOps.AssertionError();
+            throw PythonOps.AssertionError("");
```
slozier: I don't think raising an `AssertionError` with a string is technically correct, but this code path is pretty much never hit (unless you delete `__builtins__.AssertionError`), so I'm fine with this for now. I'll file an issue with the problems I found with our assert implementation.

PR #303, opened 2017-12-30
Remove BigInteger

slozier:


Src/IronPython.SQLite/Connection.cs
```diff
@@ -367,8 +367,8 @@ private static void setResult(Sqlite3.sqlite3_context ctx, object result)
                     Sqlite3.sqlite3_result_int64(ctx, (int)result);
```
gpetrou: Maybe change to a switch statement?

slozier: I thought about it but decided against since the goal of the PR is to merge in the change and not code cleanup.

PR #318, opened 2018-02-07
Add support for a chocolatey package

slide:


Build.proj
```diff
@@ -170,6 +171,13 @@
     <Delete Files="@(TempCopied)" />
   </Target>
 
+  <Target Name="ChocoPackage" Condition="'$(OS)' != 'Unix' " DependsOnTargets="Stage;MakePackageDir"
+          Outputs="$(PackageDir)\ironpython.chocolatey.$(PackageVersion).nupkg">
+    <Exec Command="$(NuGet) install chocolatey -Version $(ChocolateyVersion) -OutputDirectory &quot;$(RootDir)\packages&quot;" />
```
slozier: Any way to do this via a `PackageReference` or something like that instead in invoking nuget directly?

slide: We currently don't do any restore on Build.proj, so we would have to do that prior to invoking the build.

slozier: Don't we have the following DependsOnTargets: `ChocoPackage` -> `Stage` -> `_PlatformStage` -> `RestoreReferences`?

slide: Yes, but that restores references on the solution using the builtin Restore target for solutions. Does that automatically work for a proj?

slozier: I have no idea. I'm fine with this as it is (assuming it works). We can always change it later.

slide: I think we would have to import some .targets file or something to get a default Restore target. I tried adding a restore call, but it said there was no Restore target for Build.proj.

PR #325, opened 2018-02-14
Updated _ast.BuildAst() to prevent memory loss

lafrank:
I believe that the Parser object should be disposed after usage, thus I added a using statement to _ast.BuildAst().

### DIFF
--- a/Src/IronPython/Modules/_ast.cs
+++ b/Src/IronPython/Modules/_ast.cs
@@ -86,12 +86,13 @@ namespace IronPython.Modules
         }
 
         internal static AST BuildAst(CodeContext context, SourceUnit sourceUnit, PythonCompilerOptions opts, string mode) {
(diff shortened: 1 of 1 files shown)


slide:
Have you done any profiling to determine if this is really an area of leaking? I think we should dispose the parser, just wondering if you have any idea on what this might save

slozier:
@slide I agree that having a dispose here is correct. However, as far as I know this is on a rarely used code path which doesn't get called during normal execution so it won't lead to any savings. Even if it was called, disposing the parser only closes the underlying stream so I don't expect any memory savings.

lafrank:
Well, I am not sure how much memory is saved or wasted this way, I was only looking for memory leaks in code and identified this one as a potential one. Considering the code this is undoubtedly not the biggest source of memory leak. However, a disposable object is better to be disposed.

PR #339, opened 2018-02-22
Support new style csproj <PackageReference> usage

### LINKED ISSUE #337: IronPython.Stdlib pkg doesn't support <PackageReference />
I am trying to reference IronPython.StdLib in my csproj like:
```
    <PackageReference Include="IronPython.StdLib" Version="2.7.8"/>
```

The content files included in the package are not getting installed into my output directory.  The nuget docs say that the old style "content" folder is ignored when using <PackageReference> and a new nuget element <contentFiles> tells nuget what to do.

I think something like the following addition to the .nuspec file should correct the problem.

```
    <contentFiles>
      <files include="**" buildAction="None" copyToOutput="true" />
    </contentFiles>
```

Here are the docs that explain the changes:
https://docs.microsoft.com/en-us/nuget/reference/nuspec#specifying-files-to-include-in-the-package

hmflash:
Fixes #337 

When including StdLib with <PackageReference /> the Lib folder now shows up in the IDE and gets copied to the output directory.

### DIFF
--- a/Package/nuget/IronPython.StdLib.nuspec
+++ b/Package/nuget/IronPython.StdLib.nuspec
@@ -16,8 +16,11 @@
     <dependencies>
       <dependency id="IronPython" version="[2.7.8,3.0)" />
     </dependencies>
+    <contentFiles>
+      <files include="**\*" buildAction="None" copyToOutput="true" />
+    </contentFiles>
   </metadata>
   <files>
-    <file src="Lib\**" target="content\Lib" />
+    <file src="Lib\**" target="contentFiles\any\any\Lib" />
   </files>
 </package>


slide:
Please don't update the version in a PR. Also, are both <files><file>... entries needed, or can we just use the one that copies to contentFiles?

slide:
Go ahead and leave the version bump.

hmflash:
Naming the folder "contentFiles/any/any" is required for new style nuget.   I agree having one folder is better, I just didn't want to break current usage.  Since there was isn't currently an Install.ps1 script, it is probably safe to remove the old "content" folder.

slide:
One thing to note is that the version is overridden by the msbuild files, so bumping the version in the nuspec file won't actually do anything. So, we would have to bump the version elsewhere.

hmflash:
I can revert the version bump if you want

slide:
Yeah, lets do that and I'll determine the best way to get the new release out.

hmflash:
Version change has been removed.

slide:
Looks great, once the build checks are done, I'll take a peek at the generated nupkg file and then merge and do a release of the stdlib. Thanks for the PR!

slide:
Hmmm, closing an reopening to try and get appveyor to build

PR #368, opened 2018-05-03
Parse subject alt name on .NET Core on Linux

slozier:
Handle subject alt names properly on Linux (.NET Core).


slide:
Is there a test for this?

slozier:
I have no idea how to go about testing it.

PR #369, opened 2018-05-05
Fix #310

### LINKED ISSUE #310: exit code is not set if the code is a long
The exit code is not set if the code is a long. Note that this behavior is new with CPython 2.7.14.

Repro code:
```Python
raise SystemExit(2L)
```

slide:
This also adds a template for bug submission


slozier:
There is a test for this that should be changed:
https://github.com/IronLanguages/ironpython2/blob/55c538f9ec9a387f76fc4d114edeb9619d2df79f/Tests/test_stdconsole.py#L164

slide:
Yeah, I saw that after I created the pr and didn't get back to it today

PR #383, opened 2018-05-17
Implement pwd, grp and spwd modules

slide:
These modules will be available on Unix platforms.

Only tested on Linux for now.

test_spwd.py is not enabled to run because it required root privileges to run anything meaningful.

### DIFF
--- a/Src/IronPython.Modules/grp.cs
+++ b/Src/IronPython.Modules/grp.cs
@@ -1,0 +1,207 @@
+/* ****************************************************************************
+ *
+ * Copyright (c) Microsoft Corporation. 
+ *
+ * This source code is subject to terms and conditions of the Microsoft Public License. A 
+ * copy of the license can be found in the License.html file at the root of this distribution. If 
+ * you cannot locate the  Microsoft Public License, please send an email to 
+ * ironpy@microsoft.com. By using this source code in any fashion, you are agreeing to be bound 
+ * by the terms of the Microsoft Public License.
+ *
+ * You must not remove this notice, or any other, from this software.
+ *
+ *
+ * ***********************************************************************/
+
+#if FEATURE_NATIVE || NETCOREAPP2_0
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.IO;
+using System.IO.Pipes;
+using System.Runtime.InteropServices;
+
+using Microsoft.Scripting.Runtime;
+
+using IronPython.Runtime;
+using IronPython.Runtime.Operations;
+
+using System.Numerics;
+
+[assembly: PythonModule("grp", typeof(IronPython.Modules.PythonGrp), PythonModuleAttribute.PlatformFamily.Unix)]
+namespace IronPython.Modules {
+    
+    public static class PythonGrp {
+        public const string __doc__ = @"Access to the Unix group database.
+        
+Group entries are reported as 4-tuples containing the following fields
+from the group database, in order:
+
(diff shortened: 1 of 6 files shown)


Src/IronPython.Modules/grp.cs
```diff
@@ -0,0 +1,207 @@
...
+
+            [Documentation("group members")]
+            public List gr_mem { get; }
+
+            public override int __len__() {
+                return LENGTH;
+            }
+
+            private object[] AsArray() {
+                return new object[] { gr_name, gr_passwd, gr_gid, gr_mem };
```
slozier: Seems a bit inefficient to create an array every time we do a tuple operation? An option would be to initialize the tuple in the constructor and just make the properties index ops on the underlying array? That way we wouldn't need any of the overrides (other than `__repr__`).

Src/IronPython.Modules/nt.cs
```diff
@@ -402,18 +402,22 @@ public static class PythonNT {
         }
 
 #if FEATURE_UNIX && FEATURE_NATIVE
```
slozier: Hmm, how do we distribute the Unix build?

slide: We're not currently

slide: Travis doesn't have the capability to archive artifacts (at least last time I looked), so there isn't an easy way to distribute a release built on Travis. It would be super nice if there was a way.

Src/IronPython.Modules/grp.cs
```diff
@@ -0,0 +1,177 @@
...
+or via the object attributes as named in the above tuple.
+")]
+        public class struct_group : PythonTuple {
+
+            internal struct_group(string gr_name, string gr_passwd, int gr_gid, List gr_mem) :
+                base(new object[] { gr_name, gr_passwd, gr_gid, gr_mem }) {
+            }
+
+            [Documentation("group name")]
+            public string gr_name {
```
slozier: Nit: Maybe make these one liners with fat arrows?
```C#
public string gr_name => (string)_data[0];
```

Src/IronPython.Modules/nt.cs
slozier: Since the use of FEATURE_UNIX is very limited I wonder if we could just change them all to runtime checks and have the same binaries for all platforms. Either that or make the unix version buildable on Windows so we can include it in the generated nuget.

slide: I don't want to make Mono.Posix a reference on Windows platforms. That's my only hangup with that idea.

Src/IronPython.Modules/grp.cs
```diff
@@ -0,0 +1,161 @@
+/* ****************************************************************************
```
slozier: Maybe use the .NET Foundation copyright instead?
```
// Licensed to the .NET Foundation under one or more agreements.
// The .NET Foundation licenses this file to you under the Apache 2.0 License.
// See the LICENSE file in the project root for more information.
```

Src/IronPython.Modules/grp.cs
```diff
@@ -0,0 +1,161 @@
...
+ *
+ * You must not remove this notice, or any other, from this software.
+ *
+ *
+ * ***********************************************************************/
+
+#if FEATURE_NATIVE || NETCOREAPP2_0
+
+using System;
+using System.Collections;
```
slozier: Seems like a bunch of these are unused.

PR #396, opened 2018-05-26
ctypes fixes

### LINKED ISSUE #332: argtypes member behaves incorrectly in ctypes
Consider the following C++ code that compiles to a small library called test_lib.dll:
```cpp
#include <iostream>

#if defined(_WIN32)
  #define EXPORT __declspec(dllexport)
#else
  #define EXPORT
#endif

extern "C" 
{
    void EXPORT print_string_buffer(const char* buffer) 
    {
        std::cout << buffer << std::endl;
    }
    
    void EXPORT print_double_array(const double* array, size_t size) 
    {
        if (size == 0)
        {
            std::cout << "[]" << std::endl;
        }
        else
        {
            std::cout << "[" << array[0];
            for (size_t i = 1; i < size; i++)
            {                
                std::cout << ", " << array[i];
            }
            std::cout << "]" << std::endl;
        }
        
    }
}
```

Now consider the following two tests that call into test_lib through ctypes:
```python
import ctypes

test_lib = ctypes.CDLL("test_lib")

def test_1():    
    # test_lib.print_string_buffer.argtypes = [ctypes.c_char_p] # uncommenting this line fixes the problem
    test_lib.print_string_buffer(ctypes.create_string_buffer("1".encode()))
    
    # raises exception: _ctypes.ArgumentError: expected c_char_Array_2, got c_char_Array_3
    test_lib.print_string_buffer(ctypes.create_string_buffer("11".encode()))


def test_2():
    test_lib.print_double_array.argtypes = [ctypes.POINTER(ctypes.c_double)] # commenting this line fixes the problem
    double_array = (ctypes.c_double * 3)()
    double_array[0] = 0.1
    double_array[1] = 1.1
    double_array[2] = 2.1
    
    # raises exception: _ctypes.ArgumentError: expected c_double, got c_double_Array_3
    test_lib.print_double_array(double_array, len(double_array))
    
test_1() 
#test_2()
```

Both of these tests run fine in Python (2 and 3). However, they both fail in IronPython unless the appropriate lines are commented or uncommented as described in the tests.

For test_1, IronPython appears to be caching the argtype of the last call to print_string_buffer. If I switch the lines so that "11" prints first and "1" prints second, then the error becomes "_ctypes.ArgumentError: expected **c_char_Array_3, got c_char_Array_2**" However, the value of `print_string_buffer.argtypes` in between the two calls still shows as `None` when I print it to the console.

For test_2, IronPython is happy if I pass a single double to the function using `ctypes.byref()`, but not when I pass an array.

A google search reveals that these issues have been documented [here](https://codebright.wordpress.com/2010/06/23/strange-behaviour-using-ctypes-in-ironpython/)

slide:
Several updates for ctypes compatibility, this will eventually fix #332


Src/IronPython.Modules/_ctypes/_ctypes.cs
```diff
@@ -682,14 +690,23 @@ public static partial class CTypes {
...
             } else if (arraySize < size + offset) {
-                throw PythonOps.ValueError("Buffer size too small ({0} instead of at least {1} bytes)",
-                    arraySize,
-                    size
-                );
+                throw PythonOps.ValueError($"Buffer size too small ({arraySize} instead of at least {size} bytes)");
+            }
+        }
+
+        private static void ValidateArraySizes(BigInteger arraySize, int offset, int size) {
```
slozier: Doesn't seem to be referenced anywhere?

Src/IronPython.Modules/ModuleOps.cs
```diff
@@ -97,12 +96,17 @@ public static partial class ModuleOps {
...
             CTypes.CData res = o as CTypes.CData;
-            object asParam;
-            if (res == null && PythonOps.TryGetBoundAttr(o, "_as_parameter_", out asParam)) {
+            if (res == null && PythonOps.TryGetBoundAttr(o, "_as_parameter_", out object asParam)) {
                 res = asParam as CTypes.CData;
             }
 
-            if (res == null || res.NativeType != type) {
+            bool valid = true;
+            if(res != null && res.NativeType is CTypes.ArrayType at) {
```
slozier: Can't say this is very clear. Maybe some sort of comment would be helpful?

Src/IronPython.Modules/ModuleOps.cs
```diff
@@ -125,25 +128,26 @@ public static partial class ModuleOps {
...
                 res = asParam as CTypes.CData;
             }
-            if (res != null && res.NativeType != ((CTypes.PointerType)type)._type) {
-                if (!(res is CTypes.Pointer)) {
-                    throw ArgumentError(type, ((PythonType)((CTypes.PointerType)type)._type).Name, o);
-                }
-                return null;
+
+            bool valid = true;
+            if (res.NativeType is CTypes.ArrayType at) {
```
slozier: If `res == null` will this fail? Also as above, not clear what check is going on.

Src/IronPython.Modules/ModuleOps.cs
```diff
@@ -125,25 +128,26 @@ public static partial class ModuleOps {
...
+                valid = ((type is CTypes.ArrayType t) && (t.ElementType == at.ElementType)) ||
+                    ((type is CTypes.PointerType p) && (p._type == at.ElementType));
+            }
+
+            if (res != null && !valid) {
+                throw ArgumentError(type, ((PythonType)((CTypes.PointerType)type)._type).Name, o);
```
slozier: What if type is not CTypes.PointerType? Will the cast fail?

Src/IronPython.Modules/_ctypes/ArrayType.cs
```diff
@@ -118,37 +116,48 @@ private ArrayType(Type underlyingSystemType)
...
-                _Array res = (_Array)CreateInstance(Context.SharedContext);
+                _Array res = (_Array)CreateInstance(context);
                 res._memHolder = new MemoryHolder(((INativeType)this).Size);
                 for (int i = 0; i < ((INativeType)this).Size; i++) {
                     res._memHolder.WriteByte(i, array._bytes[i]);
                 }
                 return res;
             }
 
+            public _Array from_buffer_copy(CodeContext/*!*/ context, string data, int offset=0) {
```
slozier: Should we have some sort of check on the string to make sure it contains only bytes? I'm guessing this is only here because our strings are unicode? Maybe a comment to remove this in Py3 if this is the case.

Src/IronPython.Modules/_ctypes/Field.cs
```diff
@@ -53,7 +53,7 @@ public sealed class Field : PythonTypeDataSlot, ICodeFormattable {
                 _index = index;
                 _fieldName = fieldName;
 
-                if (bits != null) {
+                if (bits != null && bits.Value != (_fieldType.Size * 8)) {
```
slozier: Maybe some sort of comment explaining why we exclude the Size * 8 case?

Src/IronPython.Modules/_ctypes/SimpleType.cs
```diff
@@ -84,12 +86,42 @@ public SimpleType(CodeContext/*!*/ context, string name, PythonTuple bases, Pyth
...
+                _setValue = SetVal;
             }
 
             private SimpleType(Type underlyingSystemType)
                 : base(underlyingSystemType) {
             }
 
+            private SimpleType(CodeContext/*!*/ context, string name, PythonTuple bases, PythonDictionary dict, bool isLittleEndian) : this(context, name, bases, dict) {
+                _format = (isLittleEndian ? '<' : '>') + _charType.ToString();
+                if(isLittleEndian != BitConverter.IsLittleEndian) {
```
slozier: I wonder if it would be better to simply have a `swap` field and move down the swapping to the `MemoryHolder.Read/Write` methods. It would avoid using delegates and having the mostly duplicated `Get/SetValue` methods.

Src/IronPython/Runtime/Operations/ByteOps.cs
```diff
@@ -190,6 +190,8 @@ public static partial class ByteOps {
                 return ((double)(float)o).ToByteChecked();
             } else if (Converter.TryConvertToIndex(o, out i)) {
                 return i.ToByteChecked();
+            } else if(o is string str && str.Length == 1) {
+                return (byte)(str[0]);
```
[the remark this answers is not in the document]
slide: Good call

PR #412, opened 2018-06-02
Fix #380

### LINKED ISSUE #380: `*.exe` files are not created when installing packages
We recently changed our project's installation to use [setuptools entry-points](http://setuptools.readthedocs.io/en/latest/setuptools.html#automatic-script-creation) to automatically create start-up scripts, most importantly to get `*.exe` files on Windows (robotframework/robotframework#2415). Unfortunately the `*.exe` files aren't created when using IronPython 2.7.8.

slide:
Fix platform check in easy_install.py


slide:
Hmmm, not sure why my commit history is like that...

PR #415, opened 2018-06-08
Fix #375

### LINKED ISSUE #375: `UnicodeDecodeError` has no error message
When decoding bytes with bad encoding, the resulting `UnicodeDecodeError` has no error message:

```
E:\>ipy -V
IronPython 2.7.8 2.7.8.0 on .NET 4.0.30319.42000
E:\>ipy -c "b'\xe4'.decode('ASCII')"
  File "<string>", line 1, in <module>
UnicodeDecodeError
```

With CPython the message exists and also earlier IronPython versions had it.

```
E:\>python -c "b'\xe4'.decode('ASCII')"
  File "<string>", line 1, in <module>
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe4 in position 0: ordinal not in range(128)
```

Error message used to be better in earlier IronPython versions. Because exception type is correct, this isn't too severe.

slide:
Fixed error message on UnicodeDecodeError


slozier:
Hmm, does this only cover the ASCII encoding?

PR #416, opened 2018-06-12
Packaging Updates

slide:
- Move packaging targets into their own files
- Add initial deb packaging


Package/msi/Msi.Packaging.targets
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="4.0">
+  <Target Name="MsiPackage" Condition="'$(OS)' != 'Unix' " AfterTargets="Package">
+    <ItemGroup>
+      <CleanupFiles Include="$(StageDir)\net45\IronPythonTest.dll"/>
+    </ItemGroup>
+    <Delete Files="@(CleanupFiles)" />
```
slozier: Should we move this delete to the `Package` target if it's useful for all package targets?

slide: I think it's only needed for Msi because of the way it specifies files to include. The other packages allow specifying excludes and such, but the Wix stuff doesn't seem to.

slozier: Maybe I missed it, but I'm not seeing it excluded in the Debian package.

Package/nuget/NuGet.Packaging.targets
```diff
@@ -0,0 +1,33 @@
...
+    <ItemGroup>
+      <DLLsFiles Include="**\DLLs\IronPython.*.dll" />
+      <DLLsFiles Include="**\DLLs\IronPython.*.xml" />
+    </ItemGroup>
+
+    <!-- Copy the files from DLLs into the main directory for nuget packaging -->
+    <Copy SourceFiles="@(DLLsFiles)" CopiedFiles="@(CopiedFiles)" DestinationFolder="$([System.IO.Path]::GetFullPath('%(DLLsFiles.RecursiveDir)\..'))">
+      <Output TaskParameter="CopiedFiles" ItemName="TempCopied" />
+    </Copy>
+    <Exec Command="$(NuGet) pack &quot;%(NuSpecDir)\%(Package.Identity).nuspec&quot; -BasePath &quot;$(StageDir)&quot; -OutputDirectory &quot;$(PackageDir)&quot; -Version $(PackageVersion)" />
```
slozier: Could we use `dotnet pack` and get rid of the explicit nuget dependency?

slozier: nvm, looks like it's not equivalent to nuget pack...

Package/msi/Msi.Packaging.targets
slide: https://github.com/IronLanguages/ironpython2/pull/416/files#diff-ea9b249df442099c197f0858e7f580caR16

Src/IronPython/Runtime/PythonContext.cs
```diff
@@ -280,16 +280,32 @@ public PythonContext(ScriptDomainManager/*!*/ manager, IDictionary<string, objec
...
 
 #if DEBUG
                     // For developer use, add Src/StdLib/Lib
                     string devStdLib = Path.Combine(entry, @"../../../Src/StdLib/Lib");
-                    if (Directory.Exists(devStdLib))
+                    if (Directory.Exists(devStdLib)) {
                         path.append(devStdLib);
+                    }
+#else
+                    if(Environment.OSVersion.Platform == PlatformID.Unix) {
```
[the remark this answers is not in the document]
slide: That's probably a good idea.

slide:
Any other feedback on this?

PR #423, opened 2018-06-15
Add .NET Core 2.1

slozier:
Adds .NET Core 2.1 support. This also resolves https://github.com/IronLanguages/ironpython2/issues/38.

### DIFF
--- a/make.ps1
+++ b/make.ps1
@@ -98,7 +98,7 @@ function GenerateRunSettings([String] $framework, [String] $configuration, [bool
 
     $doc.AppendChild($runSettings) | Out-Null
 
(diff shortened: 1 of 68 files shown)


make.ps1
```diff
@@ -4,7 +4,7 @@ Param(
     [Parameter(Position=1)]
     [String] $target = "release",
     [String] $configuration = "Release",
-    [String[]] $frameworks=@('net45','netcoreapp2.0'),
+    [String[]] $frameworks=@('net45','netcoreapp2.1'),
```
slide: You don't want to build netcoreapp2.0 anymore?

slozier: The list of frameworks is only used for testing. We'd probably run out of time if we tested both 2.0 and 2.1.

slide: Ah, right.

Issue #420, opened 2018-06-14
Make CLR enum behave more like python IntEnum


kayoub5:
### Description

The CLR enum have interesting characteristics that makes it too similar to python `IntEnum` and `IntFlag`.
however not all of those characteristics are supported in IronPython.
Examples: given the following `DaysInt`:
```cs
[Flags]
public enum DaysInt : int {
    Mon = 0x01,
    Tue = 0x02,
    Wed = 0x04,
    Thu = 0x08,
    Fri = 0x10,
    Sat = 0x20,
    Sun = 0x40,
    Weekdays = Mon | Tue | Wed | Thu | Fri,
    Weekend = Sat | Sun,
    All = Weekdays | Weekend
  }
```
the below C# code will work just fine:
```cs
Console.WriteLine((DaysInt) 8); // Thu
Console.WriteLine((DaysInt) 0x100); // 256
Console.WriteLine(DaysInt.Tue < DaysInt.Wed); // 6
```

if `DaysInt` was implemented purely in python, it would look something like this:
```py
from enum import IntFlag

class DaysInt(IntFlag):
    Mon = 0x01,
    Tue = 0x02,
    Wed = 0x04,
    Thu = 0x08,
    Fri = 0x10,
    Sat = 0x20,
    Sun = 0x40,
    Weekdays = 0x1F,
    Weekend = 0x20 | 0x40,
    All = 0x7F,

print(DaysInt(8)) # DaysInt.Thu
print(DaysInt(0x100)) # DaysInt.256
print(DaysInt.Tue < DaysInt.Wed) # True
```
### Steps to Reproduce

Now if we take CLR implimentation of `DaysInt`, and import it to IronPython

```py
print(DaysInt(8)) # throws
print(DaysInt(0x100)) # throws
print(DaysInt.Tue < DaysInt.Wed) # throws
```
After some digging, I noticed that none of the methods to support this is implemented (`__new__`, `__gt___`,...) in `EnumOps.cs`.

I am open to creating the PR myself, if the idea itself is accepted.

### Versions

IronPython 2.7.8

kayoub5:
I know that `IntEnum` is for Python 3.4+ but it was back-ported to Python 2 as [enum34](https://libraries.io/pypi/enum34) and it's widely adopted by the Python community.

slide:
We would definitely be open to a PR, please make sure there are tests for the behavior when submitting the PR.

kayoub5:
[...]

PR #424, opened 2018-06-18
Allow creating enumerations from their numeric values

kayoub5:
Related to #420


Tests/test_enum.py
```diff
@@ -0,0 +1,56 @@
...
+
+import clr
+clr.AddReference('IronPython')
+from IronPython.Compiler.Ast import Arg
+
+Arg(None)
+
+import sys
+
+from iptest import IronPythonTestCase, is_cli, run_test, skipUnlessIronPython
```
slozier: I don't see how the code above relates to the enum tests? You can probably get rid of everything before this line and use the .NET foundation copyright.
```
# Licensed to the .NET Foundation under one or more agreements.
# The .NET Foundation licenses this file to you under the Apache 2.0 License.
# See the LICENSE file in the project root for more information.
```

Tests/test_enum.py
```diff
@@ -0,0 +1,56 @@
...
+
+if is_cli:
```
slozier: This is use for?

slozier:
So the issue I have with this approach is that you're basically injecting a fake .NET constructor instead just adding a Python constructor (for example `DaysInt.__new__(DaysInt, 8)` does not work).

Tests/test_enum.py
```diff
@@ -0,0 +1,56 @@
...
+    def setUp(self):
+        super(Enum34Test, self).setUp()
+        self.load_iron_python_test()
+    
+    @skipUnlessIronPython()
+    def test_constructor(self):
+        """check that enum could be constructed from numeric values"""
+        from IronPythonTest import DaysInt, DaysShort, DaysLong, DaysSByte, DaysByte, DaysUShort, DaysUInt, DaysULong
+        
+        enum_types = [DaysInt, DaysShort, DaysLong, DaysSByte, DaysByte, DaysUShort, DaysUInt, DaysULong]
```
slozier: Some negative tests would be useful as well. For example making sure a `ValueError` is thrown when trying to construct with invalid values.

kayoub5:
@slozier I tried `__new__`, but it never got called by Ironpython.
Those python methods are only called when explicitly invoked like what you did.

I could add `__new__` and `__call__` to `EnumOps` if that's desired.

slozier:
I guess it's probably not a big deal that `__new__` can't be invoked directly since enums can't be used as a base class anyway.

slide:
@kayoub5 Can you rebase your PR from master so that the .travis.yml configuration is correct?

kayoub5:
@slozier @slide 
Requested changes done.

Tests/test_enum.py
kayoub5: Tests added

kayoub5: Is there any configuration change required for the tests to be picked by CI?

slozier: It should be automatically picked up.

Src/IronPython/Runtime/Operations/PythonTypeOps.cs
```diff
@@ -345,13 +345,35 @@ internal static class PythonTypeOps {
...
+            }
+            return ctors;
+        }
+
+        // support for EnumType(number)
+        private static T CreateEnum<T>(object value) {
+            try {
+                return (T)Enum.ToObject(typeof(T), value);
+            } catch (ArgumentException) {
+                throw PythonOps.ValueError(
```
slide: Please use string interpolation here:

```csharp
throw PythonOps.VarlueError("{value} is not a valid {typeof(T)}")
```

Tests/test_enum.py
slide: See https://ci.appveyor.com/project/AlexEarl/ironpython2/build/2.7.8.1-603?fullLog=true#L5028

Src/IronPython/Runtime/Operations/PythonTypeOps.cs
kayoub5: You would get `null is not a valid ...` instead of `None is not valid ...`, are you sure?

slide: I was just using an example. Also, it may be better to check for null before the ToObject call. Have you tried without the try/catch and see what error is raised and shown without creating your own throw?

kayoub5: for `null` it will throw an `ArgumentNullException` which would be caught by `catch (ArgumentException)`
for the exception message, I am using the one from `enum34`, and the one from .NET contains the words `null` and `Int32`...

slide: Yes, I understand, but exceptions are slow. If you check for null at the beginning and then throw your own ValueError for that, it will be faster in that case.

kayoub5: what are the chances of someone actually doing `DaysInt(None)`?

slide: You'd be surprised!

kayoub5: done

kayoub5: anything else other than waiting for CI to pass?

slide:
Looks good, just waiting for CI

PR #435, opened 2018-07-02
Use Directory.Build.props instead of Common.proj

slozier:


slozier:
@slide Any idea why test_codecs would fail on Travis? Is there a Mono update?

PR #439, opened 2018-07-08
Buffer updates

slide:


Src/IronPython/Runtime/ByteArray.cs
```diff
@@ -904,6 +905,20 @@ public class ByteArray : IList<byte>, ICodeFormattable, IBufferProtocol
             return new ByteArray(bytes);
         }
 
+        public static ByteArray operator +(ByteArray self, string other) {
+            if(other.Any(x => x > 0xff)) {
```
slozier: `MakeByteArray` below does the same check as this. Maybe we could try catch the `MakeByteArray` and throw if necessary? Or if we don't want a try catch we could add a `TryMakeByteArray` which wouldn't throw.

Also, I wonder if we should add a note that this overload is specifically a workaround for the IPy string are unicode issue.

Src/IronPython/Runtime/Operations/StringOps.cs
```diff
@@ -164,6 +165,70 @@ public class ExtensibleString : Extensible<string>, ICodeFormattable, IStructura
...
+
+        IList<BigInteger> IBufferProtocol.GetShape(int start, int? end) {
+            if (end != null) {
+                return new[] { (BigInteger)end - start };
+            }
+            return new[] { (BigInteger)_wrapped.Length - start };
+        }
+
+        void IBufferProtocol.SetItem(int index, object value) {
+            throw PythonOps.TypeError("cannot modify read - only memory");
```
slozier: spaces: `read - only`

Src/IronPython/Runtime/PythonBuffer.cs
```diff
@@ -238,20 +223,16 @@ public PythonBuffer(CodeContext/*!*/ context, object @object, int offset)
...
-            return Math.Max(_size, 0);
+            return Math.Max(Size, 0);
         }
 
-        internal int Size {
-            get {
-                return _size;
-            }
-        }
+        internal int Size { get; private set; }
```
slozier: Maybe not worth the effort, but it seems like `_object` and `Size` should be readonly.

slozier:
Can we add some tests for all this stuff?

Src/IronPython.Modules/_socket.cs
```diff
@@ -801,10 +805,34 @@ public class socket : IWeakReferenceable {
                 + "successful completion of the Send method means that the underlying system has\n"
                 + "had room to buffer your data for a network send"
                 )]
-            public void sendall(PythonBuffer data, [DefaultParameterValue(0)] int flags) {
+            public void sendall(PythonBuffer data, int flags=0) {
                 sendallWorker(data.byteCache, flags);
             }
 
+            [Documentation("sendall(string[, flags]) -> None\n\n"
```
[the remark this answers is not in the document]
slide: I'm not sure how the doc stuff works in regards to something like help(_socket.sendall), we probably don't, but I kept it there like the other overloads.

slide:
I have another PR coming that will enable tests for this, but its part of a small rework to the testing stuff.

Src/IronPython/Runtime/Operations/StringOps.cs
slide: Thanks

PR #441, opened 2018-07-08
Fix #421, #440

### LINKED ISSUE #421: Change meaning of Condition in test case files
Currently the `Condition` in test case files is the "run condition" this is somewhat confusing when combined with the `Reason`. It seems to me like `Condition` should be the "skip condition". For example:
```
[test_argparse]
Condition='$(OS)' != 'Unix'
Reason=This fails on Linux for some reason, needs debug
```
would become
```
[test_argparse]
Condition='$(OS)' == 'Unix'
Reason=This fails on Linux for some reason, needs debug
```

slide:
Changes names of tests to make them distinguishable by path from root of tests. Also, changes `Condition` to `RunCondition`


slozier:
I assume it'll pass once you rerun the tests with the other PR merged in?

slide:
Yes, it should.

slide:
I'll rebase on master

PR #442, opened 2018-07-09
Artifact binlog on failure

slozier:


slozier:
Go figure it passes the one time I want it to fail...

PR #443, opened 2018-07-09
Fix #381

### LINKED ISSUE #381: `pip` should work after installation without needing to run `ipy -m ensurepip`
CPython Windows (and OSX) installers invoke `ensurepip` automatically as explained in [PEP-477](https://www.python.org/dev/peps/pep-0477/). IronPython Windows installer should do the same.

slide:
Add run of `ensurepip` to end of the installer


slozier:
Did you change the install folder on purpose? Won't it break VS integration?

slide:
No, I did not. That was something I didn't mean to commit. Good thing for reviews!

slozier:
Is it easy to make the ensurepip optional (but selected by default) in the installer?

slide:
Right now we don't have any component selection at all, so we would have to add that, which is not trivial as far as I remember.

slozier:
Alright. Something wrong with my installer. I'm stuck in a broken state (can't install or uninstall). Will have to look at it some more tomorrow.

PR #451, opened 2018-07-11
Fixes #21

### LINKED ISSUE #21: Strange behavior of symbolic group name on regular expression module.
_From @utgwkk on May 31, 2017 3:5_

The duplication of symbolic group names on regular expression is prohibited on CPython's `re` module.
https://docs.python.org/2/library/re.html#regular-expression-syntax 
>Group names must be valid Python identifiers, and each group name must be defined only once within a regular expression.

However, the following code returns no error on IronPython 2.7.7, and yields a strange value.
This code should be an error.

```
IronPython 2.7.7 (2.7.7.0) on .NET 4.0.30319.1 (32-bit)
Type "help", "copyright", "credits" or "license" for more information.
>>> import re
>>> ptn = re.compile(r'(?P<hoge>\w+):(?P<hoge>\w+)')
>>> ptn.match('hoge:fuga')
<RE_Match object at 0x000000000000006D>
>>> _.group('hoge')
'fuga'
```

_Copied from original issue: IronLanguages/main#1664_

slide:
Add check for named groups

### DIFF
--- a/Src/IronPython.Modules/re.cs
+++ b/Src/IronPython.Modules/re.cs
@@ -206,19 +206,11 @@ namespace IronPython.Modules {
         public class RE_Pattern : IWeakReferenceable {
             internal Regex _re;
             private PythonDictionary _groups;
-            private int _compileFlags;
+            private readonly int _compileFlags;
             private WeakRefTracker _weakRefTracker;
             internal ParsedRegex _pre;
 
-            internal RE_Pattern(CodeContext/*!*/ context, object pattern)
-                : this(context, pattern, 0) {
-            }
-
-            internal RE_Pattern(CodeContext/*!*/ context, object pattern, int flags) :
-                this(context, pattern, flags, false) {
-            }
-
-            internal RE_Pattern(CodeContext/*!*/ context, object pattern, int flags, bool compiled) {
+            internal RE_Pattern(CodeContext/*!*/ context, object pattern, int flags=0, bool compiled=false) {
                 _pre = PreParseRegex(context, ValidatePatternAsString(pattern));
                 try {
                     flags |= OptionToFlags(_pre.Options);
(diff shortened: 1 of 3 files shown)


Src/IronPython.Modules/re.cs
```diff
@@ -969,8 +962,26 @@ internal class ParsedRegex {
                                                 pattern = pattern.Substring(0, tmpIndex) + ">" + pattern.Substring(tmpIndex + 1);
                                             } else {
                                                 containsNamedGroup = true;
+                                                
+                                                // we need to look and see if the named group was already seen and throw an error if it was
+                                                if(nameIndex + 1 < pattern.Length && pattern[nameIndex + 1] == '<') {
+                                                    int tmpIndex = nameIndex + 2;
```
slozier: Isn't this just `string.IndexOf`?

slide: Yeah, just got into the mode of reusing what was above for some reason.

Src/IronPython.Modules/re.cs
```diff
@@ -960,17 +953,30 @@ internal class ParsedRegex {
...
+                                                
+                                                if (tmpIndex == -1) throw PythonExceptions.CreateThrowable(error(context), "unexpected end of regex");
 
                                                 pattern = pattern.Substring(0, tmpIndex) + ">" + pattern.Substring(tmpIndex + 1);
                                             } else {
-                                                containsNamedGroup = true;
+                                                containsNamedGroup = true;                                                
+                                                // we need to look and see if the named group was already seen and throw an error if it was
+                                                if(nameIndex + 1 < pattern.Length && pattern[nameIndex + 1] == '<') {
+                                                    int tmpIndex = pattern.IndexOf('>', nameIndex);
```
slozier: Could probably start at `nameIndex+2`, but it's a minor optimization (same thing above).

PR #452, opened 2018-07-11
Fixes #25

### LINKED ISSUE #25: Need a way to mark methods as hidden for certain OS's
_From @slide on September 20, 2016 22:22_

For instance, os.startfile should not exist on Posix OS's, only Windows.

_Copied from original issue: IronLanguages/main#1425_

slide:
Adds the ability to specify which platforms to hide a method for using the PythonHiddenAttribute.

I marked this as WIP because I would like some feedback on the naming and so forth so it's not confusing.

### DIFF
--- a/Tests/test_regressions.py
+++ b/Tests/test_regressions.py
@@ -1165,6 +1165,8 @@ class C:
         # this is not available on Linux systems
         if is_posix:
             self.assertRaises(AttributeError, lambda: os.startfile('/bin/bash'))
+        else:
+            self.assertTrue(hasattr(os, 'startfile'))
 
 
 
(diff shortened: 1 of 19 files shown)


slide:
Naming things is hard  :smile:

Tests/test_regressions.py
```diff
@@ -1158,4 +1158,14 @@ def test_ipy2_gh39(self):
         for i in rng:
             self.assertEqual(i, next(it))
 
+    @skipUnlessIronPython()
+    def test_ipy2_gh25(self):
+        """https://github.com/IronLanguages/ironpython2/issues/25"""
+
+        # this is not available on Linux systems
```
slozier: Maybe we should have the positive case as well? `hasattr(os, "startfile")`

PR #453, opened 2018-07-11
[WIP] Remove CustomTasks

slide:
This removes CustomTasks, instead relying on the tasks from MSBuildCommunityTasks that we are already using.


slozier:
What about generating a single CurrentVersion.Generated.cs instead of one per project? Maybe something to think about for the future.

slide:
That's something I was thinking as well. I'll try and work that out.

slide:
The main issue I see with creating only one file is building in Visual Studio. I could easily generate the files in Build.proj as a pre-Target to other targets, but in Visual Studio, I don't know of a way to do that because it may be that only one project is built if only one project has changes. So, in that case, do I generate it every time a project is being built, which would then cause all projects needing to be built?

PR #460, opened 2018-07-14
Fixes #370, fixes #463

### LINKED ISSUE #370: `ElementTree.iterparse` not implemented
Our project needs `ElementTree.iterparse` to parse large XML files memory efficiently. Unfortunately it is not implemented in IronPython:

```python
IronPython 2.7.8 (2.7.8.0) on .NET 4.0.30319.42000 (64-bit)
Type "help", "copyright", "credits" or "license" for more information.
>>> from xml.etree import ElementTree as ET
>>> from StringIO import StringIO
>>> ET.iterparse(StringIO('<root/>'))
  File "<stdin>", line 1, in <module>
  File "c:\Program Files\IronPython 2.7\Lib\xml\etree\ElementTree.py", line 1198, in iterparse
NotImplementedError: iterparse is not supported on IronPython. (CP #31923)
>>>
```

We have instructed our users to install [ElementTree 1.2.7 preview release from 2007](http://effbot.org/downloads/#elementtree) and it seems to work fine, but having such an extra dependency that cannot even be pip-installed is highly annoying. Incorporating the fixes from the aforementioned ElementTree version into IronPython itself would be great.

### LINKED ISSUE #463: Ironpython has problems about some python-native libs and pyd problems.
Recently, we've been working on making our old python code compatible with IronPython which uses some native python libs, such as plist-lib, sqlite3 and others. We noticed that IronPython has less efficiency than native-python code called by C/C++. Especially , the sqlite3, which would take 4 time as long to insert 1M records, compaired with code called by C++.
We are also working on handle some iOS files on PC and Plist-lib is necessary, but it will not working with IronPython as XML-lib always cause a crash. I'm wodering if there will be a way to handle those problems? and may be we can fixed it by our own under your guidance.
thanks

slide:
Adds to the initial implementation by @slozier

### DIFF
--- a/Src/IronPython.Modules/pyexpat.cs
+++ b/Src/IronPython.Modules/pyexpat.cs
@@ -1,0 +1,612 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information.
+
+using IronPython.Runtime;
+using IronPython.Runtime.Exceptions;
+using IronPython.Runtime.Operations;
+using IronPython.Runtime.Types;
+using Microsoft.Scripting;
+using Microsoft.Scripting.Runtime;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.IO;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Xml;
+
+[assembly: PythonModule("pyexpat", typeof(IronPython.Modules.PythonExpat))]
+namespace IronPython.Modules {
+    public static class PythonExpat {
+        public const int XML_PARAM_ENTITY_PARSING_NEVER = 0;
+        public const int XML_PARAM_ENTITY_PARSING_UNLESS_STANDALONE = 1;
+        public const int XML_PARAM_ENTITY_PARSING_ALWAYS = 2;
+
+        private static readonly object _errorsKey = new object();
+
+        internal enum XmlErrors : int {
+            [Description("out of memory")]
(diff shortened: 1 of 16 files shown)


Src/IronPython.Modules/pyexpat.cs
```diff
@@ -0,0 +1,608 @@
+﻿using IronPython.Runtime;
```
slozier: Maybe add the .NET Foundation header.

slozier:
Let's assume I knew what I was doing when I wrote some of that. Do we have any tests?

slozier:
We'll probably need to update StdLib.pyproj (via MakeModuleList.py) to remove the xml stuff from the blacklist.

slide:
I'm going to enable some of the test_pyexpat.py tests soon. I wanted to see how it fell out on CI

slozier:
Can we add regression tests specifically for the two issues we are closing? Something simple like:
```Python
from xml.etree import ElementTree as ET
from StringIO import StringIO
x = ET.iterparse(StringIO('<root/>'))
y = x.next()
assert y[0] == 'end' and y[1].tag == 'root'
```

```Python
import plistlib
x = '<?xml version="1.0" encoding="UTF-8"?><!DOCTYPE plist PUBLIC "-//Apple Computer//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd"><plist version="1.0"><dict><key>A</key><string>B</string></dict></plist>'
assert plistlib.readPlistFromString(x) == {'A': 'B'}
````

slide:
Sure, that's a great idea

slozier:
test_minidom is probably a low hanging fruit.

Issue #364, opened 2018-04-30
How do I install pip on ubuntu?


tzachshabtay:
Hi, downloaded the zip file from release 2.7.8.
I'm trying to install pip for IronPython on ubuntu 16.04 LTS 64 bits.

Tried the dotnet core runtime first with `bash netcoreapp2.0/ipy.sh -X Frames -m ensurepip` and got:

```
Unhandled exception:
Traceback (most recent call last):
  File "/home/tshabtay/Downloads/IronPython/Lib/runpy.py", line 187, in run_module
  File "/home/tshabtay/Downloads/IronPython/Lib/runpy.py", line 81, in _run_module_code
  File "/home/tshabtay/Downloads/IronPython/Lib/runpy.py", line 72, in _run_code
  File "/home/tshabtay/Downloads/IronPython/Lib/ensurepip/__main__.py", line 4, in <module>
  File "/home/tshabtay/Downloads/IronPython/Lib/ensurepip/__init__.py", line 231, in _main
  File "/home/tshabtay/Downloads/IronPython/Lib/ensurepip/__init__.py", line 127, in bootstrap
  File "/home/tshabtay/Downloads/IronPython/Lib/ensurepip/__init__.py", line 47, in _run_pip
  File "/tmp/tmpWUMHbt/pip-8.1.1-py2.py3-none-any.whl/pip/__init__.py", line 16, in <module>
  File "/tmp/tmpWUMHbt/pip-8.1.1-py2.py3-none-any.whl/pip/vcs/subversion.py", line 9, in <module>
  File "/tmp/tmpWUMHbt/pip-8.1.1-py2.py3-none-any.whl/pip/index.py", line 30, in <module>
  File "/tmp/tmpWUMHbt/pip-8.1.1-py2.py3-none-any.whl/pip/wheel.py", line 32, in <module>
  File "/tmp/tmpWUMHbt/pip-8.1.1-py2.py3-none-any.whl/pip/pep425tags.py", line 9, in <module>
  File "/home/tshabtay/Downloads/IronPython/Lib/ctypes/__init__.py", line 7, in <module>
ImportError: No module named _ctypes
```

That's after I've set  `IRONPYTHONPATH` to point at Lib.

Then I tried mono with net40 with `mono net40/ipy.exe -X Frames -m ensurepip` and got:

```
Unhandled exception:
Traceback (most recent call last):
TypeError: Value cannot be null.
Parameter name: method
```

[...]

PR #465, opened 2018-07-19
[WIP] Attempt to get Mono.Posix working with zip package

slozier:
So the idea is to place the `runtimes` folder of the `Mono.Posix.NETStandard` package in the `netcoreapp2.0` and `netcoreapp2.1` folders of the zip package and then load them up using a deps file.

Maybe this isn't the way to go... Perhaps we should be doing a `dotnet publish` and including the `runtimes` folder as well as `ipy.deps.json` with the application?

@slide The part I haven't figured out it how to get the `runtimes` folder from the nuget package to the staging folder so it can get packaged.

Related to https://github.com/IronLanguages/ironpython2/issues/364

### DIFF
--- a/Src/IronPython/Runtime/PythonContext.cs
+++ b/Src/IronPython/Runtime/PythonContext.cs
@@ -287,7 +287,7 @@ namespace IronPython.Runtime
                         path.append(devStdLib);
                     }
 #else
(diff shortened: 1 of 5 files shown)


slozier:
Closing this. Will revisit later.

Issue #306, opened 2018-01-09
help not finding documentation for .NET Framework reference assemblies


slozier:
It appears that the documentation does not load properly (on my machine). When running the repro code, the `help` method will try to load the documentation from `C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.0\mscorlib.xml` however this file appears to just be a stub redirecting to `C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.X\mscorlib.xml`. Since the code does not follow these redirects it results in no documentation loading.

I'm not what piece of software results in the stub docu files being installed...

Repro code:
```Python
import System
help(System.DateTime.Parse)
```

PR #466, opened 2018-07-20
Attempt to follow redirects in XML doc files

slozier:
Resolves https://github.com/IronLanguages/ironpython2/issues/306

### DIFF
--- a/Src/IronPython/Runtime/Types/DocBuilder.cs
+++ b/Src/IronPython/Runtime/Types/DocBuilder.cs
@@ -599,6 +599,15 @@ namespace IronPython.Runtime.Types {
                     xpd = _CachedDoc;
                 } else {
                     xpd = new XPathDocument(xml);
+                    // attempt to redirect if required
+                    var node = xpd.CreateNavigator().SelectSingleNode("/doc/@redirect");
+                    if (node != null) {
+                        var redirect = node.Value;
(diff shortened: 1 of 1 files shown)


Src/IronPython/Runtime/Types/DocBuilder.cs
```diff
@@ -599,6 +602,17 @@ static class DocBuilder {
                     xpd = _CachedDoc;
                 } else {
                     xpd = new XPathDocument(xml);
+                    if (isRef) {
+                        // attempt to redirect if required
+                        var node = xpd.CreateNavigator().SelectSingleNode("/doc/@redirect");
+                        if (node != null) {
+                            var redirect = node.Value;
+                            redirect = redirect.Replace("%PROGRAMFILESDIR%", Environment.GetFolderPath(Environment.Is64BitProcess ? Environment.SpecialFolder.ProgramFilesX86 : Environment.SpecialFolder.ProgramFiles) + Path.DirectorySeparatorChar);
```
slide: Can you use Environment.ExpandEnvironmentVariables? https://msdn.microsoft.com/en-us/library/system.environment.expandenvironmentvariables%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396

Not sure if it takes into account the 64-bit vs. 32-bit

slozier: Unfortunately not because it's not actually an environment variable...

slide:
FYI, I ran this manually on my local clone and it fixes the issue for me.

PR #474, opened 2018-07-26
Add ability to set a number of times to retry running a test

slide:


Src/IronPythonTest/Cases/CaseExecuter.cs
```diff
@@ -90,6 +90,30 @@ class CaseExecuter {
         }
 
         public int RunTest(TestInfo testcase) {
+            int retryCount = testcase.Options.RetryCount;
+            if(retryCount > 0) {
+                int res = -1;
+                for(int i = 0; i < retryCount; i++) {
+                    try {
+                        res = RunTestImpl(testcase);
```
slozier: Is the loop break on success missing? Maybe I'm just tired.

slide: Haha, nope. Good catch!

slide: I meant, yes, it is missing.

PR #477, opened 2018-07-31
Fix #381

### LINKED ISSUE #381: `pip` should work after installation without needing to run `ipy -m ensurepip`

slide:
Add a feature for pip (selected by default) which will run ensurepip after files are copied. The uninstall will also run the ensurepip._uninstall.


slozier:
I'm sure I've asked this before. Can't we pull WiX from NuGet?

slide:
We could, yes. I don't see why it would be that much better right now though.

slozier:
Would keep the size of the repo down. For example, if I download the source for this PR, it's an 5MB of (compressed) data.

slide:
It wouldn't help for cloning since even if you delete the item, its still in the history. We'd have to delete from the history as well.

slozier:
Yes it would help with cloning. When cloning you're basically downloading the full history, so we're downloading what's in there right now and then a second copy of the binaries from the updates in this PR.

Issue #476, opened 2018-07-31
IronPython DLL compilation in Linux doesn't work with modules in sub-folders


dracan:
Hi. I'm having an issue building an IronPython DLL from Python scripts in Linux when the Python script references a module in a sub-directory.

I've posted about this on Stackoverflow, so rather than me repeating it here, I'll just link to it ...
https://stackoverflow.com/questions/51612887/ironpython-dll-compilation-in-linux-doesnt-work-with-modules-in-sub-folders

slozier:
Confirming that it's a path separator issue.

dracan:
Thanks @slozier - what compilation are you referring to? IronPython itself or my building of the DLL? In my case, I'm building on Linux, but want to use it in Windows.

slozier:
I mean `clr.CompileModules`. When you run it on Linux it embeds `/` characters, but when you run on Windows it embeds `\` characters. When loading the DLL on Linux is looks for `/` characters and on Windows it looks for `\` characters. This means that if the DLL is generated on Linux it only works on Linux and if it's generated on Windows it only works on Windows. Unfortunately it's not great for portability.

We should probably normalize it to use the `/` character which is also a valid separator on Windows.

slozier:
By adding the following to `CompiledLoader.AddScriptCode` it should be able to load a DLL generated on Linux.
```C#
name = name.Replace(Path.AltDirectorySeparatorChar, '.');
```

dracan:
Wow, that sounds as fast! Thank you, I'll give it a go on Friday when I'm back in the office :-)

PR #478, opened 2018-07-31
Path fix for loading compiled modules cross-platorm

slozier:
Resolves https://github.com/IronLanguages/ironpython2/issues/476


slozier:
Heh, a different unrelated test failed each time.

PR #480, opened 2018-07-31
[WIP] Initial macOS package support

slide:
Initial support for building a package for macOS

### DIFF
--- a/Src/IronPython/Runtime/PythonContext.cs
+++ b/Src/IronPython/Runtime/PythonContext.cs
@@ -288,12 +288,23 @@ namespace IronPython.Runtime
                     }
 #else
                     if(!IronPython.Runtime.ClrModule.IsNetCoreApp && Environment.OSVersion.Platform == PlatformID.Unix) {
-                        if(Directory.Exists("/usr/lib/ironpython2.7")) {
-                            path.append("/usr/lib/ironpython2.7");
-                        }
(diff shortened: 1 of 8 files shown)


Src/IronPython/Runtime/PythonContext.cs
```diff
@@ -288,12 +288,23 @@ public PythonContext(ScriptDomainManager/*!*/ manager, IDictionary<string, objec
                     }
 #else
                     if(!IronPython.Runtime.ClrModule.IsNetCoreApp && Environment.OSVersion.Platform == PlatformID.Unix) {
-                        if(Directory.Exists("/usr/lib/ironpython2.7")) {
-                            path.append("/usr/lib/ironpython2.7");
-                        }
+                        if(IronPython.Runtime.ClrModule.IsMacOS) {
```
slozier: What about using `if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))` and `else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux)` and getting rid of `Environment.OSVersion.Platform == PlatformID.Unix`? We could probably get rid of the `IsNetCoreApp` check as well, not sure that it's all that useful.

slide: Yeah, that's a good idea. I keep forgetting about RuntimeInformation.

PR #485, opened 2018-08-03
Fix #34, fix #482

### LINKED ISSUE #34: `__eq__` of OrderedDict subclasses does not have precedence when comparing with OrderedDicts
_From @pekkaklarck on February 19, 2015 11:37_

First of all, is this the correct place to submit IronPython bugs or should I still use the tracker at http://ironpython.codeplex.com/? There aforementioned page has a note that source is here on GitHub but doesn't tell anything about issue tracker.

---

Running this code on CPython succeeds but with IronPython the second assert fails:

``` python
from collections import OrderedDict

class Example(OrderedDict):

    def __eq__(self, other):
        return True

e = Example()
o = OrderedDict(a=1)

assert e == o, 'e != o'
assert o == e, 'o != e'
```

Because `Example` is a subclass of `OrderedDict`, its `__eq__` should be used in both of the above comparisons.

Interestingly this version with `int` works correctly:

``` python
class Example(int):

    def __eq__(self, other):
        return True

e = Example()
o = 42

assert e == o, 'e != o'
assert o == e, 'o != e'
```

_Copied from original issue: IronLanguages/main#1168_

### LINKED ISSUE #482: `__eq__` of str subclasses does not have precedence when comparing with strings
The following test case fails:
```Python
class test(str):
    def __eq__(self, other):
        return True

assert "a" == test("b")
```

May be related to https://github.com/IronLanguages/ironpython2/issues/34

slide:
Modify fix from old repo that was not integrated

### DIFF
--- a/Src/IronPython/Runtime/Binding/PythonProtocol.Operations.cs
+++ b/Src/IronPython/Runtime/Binding/PythonProtocol.Operations.cs
@@ -1132,11 +1132,11 @@ namespace IronPython.Runtime.Binding {
             bool shouldWarn = false;
 
             // if the left operand is an instance of a built-in type or a new-style class, and the right operand is an instance of a proper subclass of that type or class
-            // and overrides the base’s __rop__() method, the right operand’s __rop__() method is tried before the left operand’s __op__() method.
+            // and overrides the base's __rop__() method, the right operand's __rop__() method is tried before the left operand's __op__() method.
             PythonType xPythonType = MetaPythonObject.GetPythonType(xType);
             PythonType yPythonType = MetaPythonObject.GetPythonType(yType);
             bool opReverse = false;
-            if ((xPythonType.IsSystemType || !xPythonType.IsOldClass) && (yPythonType != xPythonType) && BindingHelpers.IsSubclassOf(yType, xType) && (rop != null)) {
+            if ((rop != null) && (xPythonType.IsSystemType || !xPythonType.IsOldClass) && (yPythonType != xPythonType) && BindingHelpers.IsSubclassOf(yType, xType)) {
                 SlotOrFunction tmp = rop;
                 rop = fop;
                 fop = tmp;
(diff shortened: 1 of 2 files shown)


Src/IronPython/Runtime/Binding/PythonProtocol.Operations.cs
```diff
@@ -1130,13 +1130,24 @@ static partial class PythonProtocol {
...
 
             bool shouldWarn = false;
-            WarningInfo info = null;
+
+            // if the left operand is an instance of a built-in type or a new-style class, and the right operand is an instance of a proper subclass of that type or class
+            // and overrides the base’s __rop__() method, the right operand’s __rop__() method is tried before the left operand’s __op__() method.
+            PythonType xPythonType = MetaPythonObject.GetPythonType(xType);
+            PythonType yPythonType = MetaPythonObject.GetPythonType(yType);
+            bool opReverse = false;
+            if ((xPythonType.IsSystemType || !xPythonType.IsOldClass) && (yPythonType != xPythonType) && BindingHelpers.IsSubclassOf(yType, xType) && (rop != null)) {
```
slozier: Minor optimization: `rop != null` check first so we don't have to check everything else

Src/IronPython/Runtime/Binding/PythonProtocol.Operations.cs
```diff
@@ -1130,13 +1130,24 @@ static partial class PythonProtocol {
             SlotOrFunction.GetCombinedTargets(cmp, rcmp, out cmp, out rcmp);
 
             bool shouldWarn = false;
-            WarningInfo info = null;
+
+            // if the left operand is an instance of a built-in type or a new-style class, and the right operand is an instance of a proper subclass of that type or class
+            // and overrides the base’s __rop__() method, the right operand’s __rop__() method is tried before the left operand’s __op__() method.
```
slozier: Fix the quotes to use `'` instead?

slozier:
Any idea what the issues were with the original PR?

slide:
One of the main issues is that IsSubClassOf returns true even if the types are the same. This is why I added the check to see if the PythonTypes were not the same.

Src/IronPython/Runtime/Binding/PythonProtocol.Operations.cs
slide: Copy and paste issue from the original PR

Src/IronPython/Runtime/Binding/PythonProtocol.Operations.cs
slide: Will do

Issue #382, opened 2018-05-15
Newer versions of pip (than bundled in 2.7.8) fail because of pyparsing issue


slide:
### Prerequisites

* [x] Are you running the latest version?
* [x] Are you reporting to the correct repository?
* [x] Did you perform a cursory search?

### Description

When trying to use pip newer than the version supplied with 2.7.8, there is an issue because pyparsing is using f_back and so forth to determine function names. This is not working in IronPython.

### Steps to Reproduce

1. ipy -m ensurepip
2. ipy -m pip install pyserial

**Expected behavior:** 

Pip should function correctly

**Actual behavior:** 

Error about -1 index

### Versions

2.7.8

littleq0903:
**Workaround included**

This is happened not only in pip, also in setuptools or any libs using pyparsing.

It's due to that traceback lib in IronPython could not extract out any stack information (returned empty list instead)

Reference here, also workaround included:
https://github.com/pyparsing/pyparsing/issues/2

The problematic part is in util function which trim the arguments number to specific number, it should be fine with modifying the code directly as a workaround.

Once this got fixed, need to replace the pyparsing version under those libs as well to fix this up.

littleq0903:
Thanks for the quick fix!

PR #489, opened 2018-08-03
Add function stack to traceback

slozier:
Resolves https://github.com/IronLanguages/ironpython2/issues/382


slide:
Yes, we will be doing an RC soon and then a final release a little later.

PeterSorrell:
Thanks - I look forward to it.

PR #490, opened 2018-08-05
Allow creating enum from their names.

kayoub5:
Related to #420


kayoub5:
@slide could this make it into the `2.7.9` release?

PR #531, opened 2018-10-03
Fix #525

### LINKED ISSUE #525: IronPython.StdLib 2.7.8.1 Nuget Package not creating a content directory
### Description

Installing IronPython.StdLib 2.7.8.1 using Nuget will not pull in the Lib to the project. The Nuget package for 2.7.8.1 only creates a contentFiles directory in the package. From what i understand only .Net Core use contentFiles. For a Nuget package to support both .Net Framework and .Net Core it needs to create a contentFiles and a content directory.

### Steps to Reproduce

1. Create a new project targeting .NetFrameWork 4.6.2
2. Install the Nuget Package IronPython.StdLib 2.7.8.1
3. Compile

**Expected behavior:** 
A Lib folder should be in the project and should be included in the output when compiling the project.

**Actual behavior:**
There is no Lib folder and nothing is included when compiling the project.

### Versions
 IronPython.StdLib 2.7.8.1
 .Net Framework 4.6.2

slide:
Add content directory for projects that use packages.config

### DIFF
--- a/Package/nuget/IronPython.StdLib.nuspec
+++ b/Package/nuget/IronPython.StdLib.nuspec
@@ -22,5 +22,6 @@
   </metadata>
   <files>
     <file src="Lib\**" target="contentFiles\any\any\Lib" />
+    <file src="Lib\**" target="content\Lib" />
   </files>
 </package>


slozier:
Is there a way to test this without publishing to NuGet?

slide:
Yes, you can add a local directory to the sources list for nuget when installing a package (or via the settings in VS). Something like this:

```
nuget install IronPython.StdLib -Source "\\path\to\directory\with\nupkg;https://api.nuget.org/v3/index.json" 
```

slide:
Yeah, I've read some interesting threads about this change

PR #532, opened 2018-10-05
Allow using mmap with io.opened file

slozier:
Resolves https://github.com/IronLanguages/main/issues/1081

This is basically more duct tape on the file manager in order to get mmap to work with io.open.


slide:
Looks like the PythonType attribute isn't working correctly on the MmapUnix class in mmap.cs. If you `import mmap` and then `dir(mmap)` on Linux, it shows MmapUnix instead of mmap.

slozier:
Probably because we have multiple types with it? Maybe we can stick it on the constructor instead?

slide:
Only interfaces, classes and structs can have the PythonTypeAttribute

PR #534, opened 2018-10-08
Switch to getter for mmap module member

slide:


slide:
The wrong stuff, but it actually doesn't match CPython already, I think I can fix it to match though

PR #548, opened 2018-10-30
Use py3 compatible syntax in tests

slozier:
https://github.com/IronLanguages/ironpython3/pull/477

### DIFF
--- a/Tests/interop/net/dynamic/test_dynamic_regressions.py
+++ b/Tests/interop/net/dynamic/test_dynamic_regressions.py
@@ -2,10 +2,9 @@
 # The .NET Foundation licenses this file to you under the Apache 2.0 License.
 # See the LICENSE file in the project root for more information.
 
-
 """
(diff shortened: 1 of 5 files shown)


equiclouds:
@slide 

I installed ironpython2.7.9 on rasperberry, and code bellow:

class test:
    def extend(self, a, *b):
        return a, b
def main():
    c1 = test().extend({},{})
    for i in range(0,100):
        c2 = test().extend({},{})
        if c1!=c2:
            return i,c1,c2
main()

Randomly, after some times, the extend function b paramters will be valued with test() object...

```
root@raspberrypi:/tmp/tmp# ipy test.ipy 
(39, ({}, ({},)), ({}, <__main__.test instance at 0x000000000000006C>))
root@raspberrypi:/tmp/tmp# ipy test.ipy 
(82, ({}, ({},)), ({}, <__main__.test instance at 0x000000000000006C>))
```

slide:
@equiclouds Please don't hijack pull requests to bring up issues that don't relate at all.

PR #560, opened 2018-12-04
Fixes #541

### LINKED ISSUE #541: dictproxy should not be hashable
```Python
class C(object): pass

hash(C.__dict__)
```
should throw
> TypeError: unhashable type: 'dictproxy'

The solution is basically to set `dictproxy.__hash__` to `None`.

slide:


slozier:
I probably should have noted where I ran into this issue...

An oddity to note, with CPython:
```Python
x = type.__dict__
x.__hash__() # this works
hash(x) # this throws
```

Issue #571, opened 2019-01-02
IronPython.Compiler.Ast.PythonVariable is not public type


lafrank:
I have an application that makes extensive use of PythonAst to create auto-completion lists in code editor and I have a lot of trouble because PythonVariable type is not public.

For instance, I would like to retrieve the list of class variables from ClassDefinition.Body.Variables but this is a type of Dictionary<string, PythonVariable> where PythonVariable is inaccessible due to its protection level. There is a workaround though, using reflection and some dynamic object inspection but it is truly ugly and most probably slow. I know I could fork IronPython and make this change myself but I am not too keen creating a private IronPython version to distribute with my application.

Therefore, I would really appreciate if you could make aforementioned type public. unless there is a very good reason to keep it hidden.

Thank you.

simplicbe:
You could make a pull request to see what changes are required. Further discussions could be placed there.

slide:
This is fixed now via #573

PR #573, opened 2019-01-03
Updated _ast.BuildAst() to prevent memory loss

lafrank:
Free up unused Parser object


simplicbe:
From my point of view, this looks good. @lafrank is your code completion public?

simplicbe:
Related to #571

lafrank:
Code completion is not yet stable and concise, hence not public. Once
finished, I will consider publishing the code.

PR #574, opened 2019-01-03
 Made PythonVariable class public

lafrank:
I have an application that makes extensive use of PythonAst to create auto-completion lists. It would be very beneficial to have IronPython.Compiler.Ast.PythonVariable class as public, otherwise it is troublesome parsing AST structure.


slide:
This was already incorporated in your other PR (https://github.com/IronLanguages/ironpython2/pull/573)

PR #575, opened 2019-01-03
Remove dependency on MSBuild Community Tasks

slide:
Removes dependency on MSBuild Community Tasks with inline tasks for
FileUpdate and Zip.


Build/Tasks.Targets
```diff
@@ -0,0 +1,75 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="4.0">
+  <UsingTask TaskName="Zip" TaskFactory="CodeTaskFactory" AssemblyFile="$(MSBuildToolsPath)\Microsoft.Build.Tasks.v4.0.dll">
```
slozier: I'm guessing this will fail with the .NET Core msbuild (`dotnet build Build.proj /t:Stage`)? I wonder if we could use https://www.nuget.org/packages/Microsoft.Build.Tasks.Core instead.

slide: I ran that command and it didn't fail, though it didn't run the staging it looks like...

slide: I'm not going to support the staging and release process without mono for now, so I don't think this is an issue.

PR #580, opened 2019-01-23
Add platform checks

slozier:


Src/IronPython.Modules/nt.cs
```diff
@@ -1549,16 +1549,26 @@ public class stat_result : IList, IList<object> {
...
 
-            //If the calls to GenerateConsoleCtrlEvent didn't work, simply
-            //forcefully kill the process.
-            Process toKill = Process.GetProcessById(pid);
-            toKill.Kill();
-            return;
+                // If the calls to GenerateConsoleCtrlEvent didn't work, simply
+                // forcefully kill the process.
+                Process toKill = Process.GetProcessById(pid);
+                toKill.Kill();
```
slide: Do we need to check for null here just in case? I've seen `GetProcessById` return null before, even when the process was alive. The user may pass in an invalid pid though a well.

slozier: Do you remember on what platform it was returning `null`? I just checked the reference source for .NET Framework, .NET Core and Mono and it either throws or returns a `new Process` so it can't be `null`. Maybe you're thinking of something else?

slide: It was on .NET desktop, but it may have been several versions ago. I have checks in an app I wrote specifically for that reason.

slozier: I just checked the .NET 4.5.1 reference source and it also can't return `null`. Maybe it was an older version which we don't support anymore.

Either way, I'm fine with this throwing if it's `null`. Better than silently not doing anything.

PR #583, opened 2019-01-24
Enable ctypes tests on .NET Core

slozier:


slozier:
Hmm, I think the failures are because Mono.Posix.NETStandard isn't working properly. As much as I don't like having to P/Invoke I'm not sure I like having the dependency.

slozier:
Ah nvm my last comment, the missing methods are in a FEATURE_NATIVE conditional which doesn't compile with .NET Core. I'll try adding FEATURE_NATIVE to .NET Core to see if it help...

PR #585, opened 2019-01-25
Fix #584

### LINKED ISSUE #584: __getattr__ incorrectly gets called when binding to user type instance
As discussed with @geekbozu on gitter:
```Python
class NoValue(object):
    def __getattr__(self, attr):
        raise Exception(attr)

noValue = NoValue()

class test(object):
    defaultValue = noValue

assert test().defaultValue is noValue
```

This is causing failures with pyasn1.

slozier:


Src/IronPython/Runtime/Types/PythonTypeUserDescriptorSlot.cs
```diff
@@ -74,17 +73,11 @@ public sealed class PythonTypeUserDescriptorSlot : PythonTypeSlot {
         }
 
         internal override bool IsSetDescriptor(CodeContext context, PythonType owner) {
-            object dummy;
-            return PythonOps.TryGetBoundAttr(context, Value, "__set__", out dummy);
+            PythonType pt = DynamicHelpers.GetPythonType(Value);
+            if (pt.IsSystemType) return false;
+            return pt.TryResolveSlot(context, "__set__", out _);
```
slide: I haven't looked at this for time, but why the switch from TryGetBoundAttr to TryResolveSlot?

PR #599, opened 2019-02-20
Support for .Net Standard

YairHalberstadt:
I think this PR will need someone to review the places where I've decided to, or not to, make compilation conditional on NETSTANDARD2_0 being, or not being, defined. All the cases where I've made compilation conditional will be visible in the PR. However of the cases where I haven't made any changes, the one I am least sure about is IronPython.Runtime.Types.DocBuilder.cs line 576.

### DIFF
--- a/Build/netstandard2.0.props
+++ b/Build/netstandard2.0.props
@@ -1,0 +1,51 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <IsFullFramework>false</IsFullFramework>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <Features>$(Features);FEATURE_APARTMENTSTATE</Features>
+    <Features>$(Features);FEATURE_APPLICATIONEXCEPTION</Features>
+    <Features>$(Features);FEATURE_ASSEMBLY_CODEBASE</Features>
+    <Features>$(Features);FEATURE_ASSEMBLY_LOCATION</Features>
+    <Features>$(Features);FEATURE_ASSEMBLY_RESOLVE</Features>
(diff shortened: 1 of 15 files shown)


slozier:
@YairHalberstadt I would probably exclude it in DocBuilder as well since having .NET Framework documentation doesn't necessarily make sense using .NET Standard assemblies, however I'm not too worried about it. Most users don't even have the reference assemblies installed on the computers.

YairHalberstadt:
I've changed DocBuilder now

YairHalberstadt:
I've now tested creating a nuget package on alocal feed, and using it to run python plugins from a .Net Standard project in our companies main solution. Everything works as expected.

slozier:
Great, thanks for the PR!

YairHalberstadt:
Is it possible to deploy a new budget package with these changes please?

slozier:
My only concern at this point (well, other than figuring out how to do a release) is the prerelease package dependency which produces the following warning:
> EXEC : warning : Issue found with package 'DynamicLanguageRuntime'.
EXEC : warning : Issue: Prerelease dependency in stable package.
EXEC : warning : Description: A stable release of a package should not have a prerelease dependency.
EXEC : warning : Solution: Either modify the version spec of dependency "System.Reflection.Emit [4.6.0-preview.19073.11, )" or update the version field in the nuspec.

Can this cause issues to people trying to consume the package?

YairHalberstadt:
I don't believe so, but it could be published as a prerelease package.

YairHalberstadt:
Which might be a good idea anyway, as it will give a chance to see if any issues occur in wild.

slozier:
I was thinking about that, but then we'd be stuck in prerelease mode until the stable RefEmit is released (alongside .NET Core 3.0 I assume). Not really an issue I guess since historically IronPython has been quite slow to release.

YairHalberstadt:
Alternatively we could use the stable versions of the reflection.Emit package. Everything *should hopefully* work, but the error message when used on AOT would be less helpful.

PR #636, opened 2019-05-28
Add support for WPF on .NET Core 3.0

slozier:


slozier:
Closing this since it looks like Microsoft won't allow cross-compilation of `Microsoft.NET.Sdk.WindowsDesktop` in the 3.0 timeframe. Also there are no NuGet packages to reference the required assemblies directly.

PR #640, opened 2019-06-05
Fix #639

### LINKED ISSUE #639: Issue with setting tracing and instantiating a class - Unable to cast object of type Func<PythonFunction,Object> to Func<CodeContext, CodeContext>
### Description
Originally posted to StackOverflow here:
https://stackoverflow.com/questions/56380832/ironpython-is-throwing-the-following-exception-unable-to-cast-object-of-type-fu

There appears to be an issue with setting tracing "on the fly" and then instantiating a class.

### Reproducing
I put up a branch with a simple console app that reproduces the issue here:
https://github.com/jacks-bush/ironpython2/tree/ReproduceCastingBug

Code for app is as follows:
```c#
using IronPython.Hosting;
using Microsoft.Scripting.Hosting;
using System.Collections.Generic;

namespace ReproduceError {
    class Program {
        static void Main(string[] args) {
            string mainCode = @"
import sys

def test(frame, event, args):
    pass

sys.settrace(test)

class BasicClass:
    pass

BasicClass()
";

            ScriptEngine engine = Python.CreateEngine(new Dictionary<string, object>() { { "CompilationThreshold", 0 } });
            ScriptScope scope = engine.CreateScope();
            engine.Execute(mainCode, scope);
        }
    }
```
### Versions
Bug has been confirmed on the latest version at the time of writing.

slozier:
Resolves https://github.com/IronLanguages/ironpython2/issues/639

@jacks-bush Is this something you can test with your application? The fix is in https://github.com/IronLanguages/ironpython2/commit/2e8aa226427ea602f6f1567462ad86cda5641fa5 the rest is just cleanup.

### DIFF
--- a/Src/IronPython/Runtime/FunctionCode.cs
+++ b/Src/IronPython/Runtime/FunctionCode.cs
@@ -717,6 +717,7 @@ namespace IronPython.Runtime {
             if (context.EnableTracing && _lambda != null) {
                 if (_tracingLambda == null) {
                     if (!forceCreation) {
+                        if (_lambda is Compiler.Ast.ClassDefinition) return; // ClassDefinition does not have the same signature as call targets
                         // the user just called sys.settrace(), don't force re-compilation of every method in the system.  Instead
                         // we'll just re-compile them as they're invoked.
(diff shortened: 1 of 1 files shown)


jacks-bush:
@slozier I think this is something I should be able to test with our application.

jacks-bush:
@slozier This fix appears to work with our software.

PR #674, opened 2019-10-09
Use runpy.py and __future__.py from StdLib

slozier:


slozier:
Equivalent of https://github.com/IronLanguages/ironpython3/pull/328

PR #705, opened 2020-04-27
Update StdLib to 2.7.18

slozier:


slozier:
Not merging this yet as it breaks pip.

PR #727, opened 2020-08-09
Deterministic build

slozier:


slozier:
@BCSharp I know you're mainly focused on ipy3, but any idea what might be going on with the macOS build? I don't have access to a mac to try myself. Same PR worked with on ipy3 (https://github.com/IronLanguages/ironpython3/pull/924).

BCSharp:
@slozier  I don't think the failures are caused by the changes from this specific PR. When I simply check out master from ironpython2 on macOS, I get the same failures.

Possibly I will have more time to investigate further in the coming days.

PR #797, opened 2021-10-15
Use Mono.Unix for .NET Core runtime

### LINKED ISSUE #795: Use Mono.Unix for Apple Silicon support
### Description

Currently IronPython requires [Mono.Posix.NetStandard](https://www.nuget.org/packages/Mono.Posix.NETStandard/5.20.1-preview), at least when running on .NET 5/6 on macOS.  However, that library only supports x64 and x86.   One possible solution is to use [Mono.Unix](https://www.nuget.org/packages/Mono.Unix) instead, which supports many more target architectures (including arm64).  Another option is to remove this dependency altogether.

On a related note, it would be nice if the dependency was added to IronPython's nuget package and/or documentation since it is required for proper execution.

### Steps to Reproduce

1. Execute a script with `os.path.exists('/path/to/my/file.txt')` without the Mono.Posix.Netstandard package installed

**Expected behavior:** 

Should return `true` if the file actually exists

**Actual behavior:** 

Returns `false`.

### Versions

IronPython 2.7.11

cwensley:
This brings in support for running on Apple Silicon on macOS, as well as iOS, android, maccatalyst, etc.

Fixes #795

### DIFF
--- a/Src/IronPython.Modules/IronPython.Modules.csproj
+++ b/Src/IronPython.Modules/IronPython.Modules.csproj
@@ -14,7 +14,7 @@
     </Reference>
   </ItemGroup>
   <ItemGroup Condition=" '$(IsFullFramework)' != 'true' ">
-    <PackageReference Include="Mono.Posix.NETStandard" Version="1.0.0" />
+    <PackageReference Include="Mono.Unix" Version="7.1.0-final.1.21458.1" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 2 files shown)


slozier:
Looks like the same change is causing a failure on macOS in IronPython 3 (https://github.com/IronLanguages/ironpython3/pull/1301). I'm guessing it's a big with the package (since it was working with the netstandard version) but unfortunately I don't have access to a mac to try and pinpoint the issue.

PR #827, opened 2023-01-12
Fix CVE-2015-20107

AkanshDivker:
Implemented fix from python/cpython#91993 for CVE-2015-20107 in mailcap.py

### DIFF
--- a/Src/StdLib/Lib/mailcap.py
+++ b/Src/StdLib/Lib/mailcap.py
@@ -1,8 +1,15 @@
 """Mailcap file handling.  See RFC 1524."""
 
 import os
+import warnings
+import re
 
 __all__ = ["getcaps","findmatch"]
+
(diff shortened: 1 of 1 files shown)


Src/StdLib/Lib/mailcap.py
```diff
@@ -1,9 +1,16 @@
 """Mailcap file handling.  See RFC 1524."""
 
 import os
+import warnings
+import re
 
 __all__ = ["getcaps","findmatch"]
 
+_find_unsafe = re.compile(r'[^\xa1-\U0010FFFF\w@+=:,./-]').search
```
slozier, line 9: This regex won't work with Python 2.

AkanshDivker: Apologies, I missed that. I believe that should be resolved now with the updated regex.

slozier:
Thanks!

PR #849, opened 2026-07-06
build for net8 and net10

BradBarnich:


BradBarnich:
sorry, misclick, meant to open on fork

